PR #8, opened 2016-12-30
Docker support

noway421:
Using `python:2.7-alpine` base which is built on top of alpine linux. Targeted towards production deployments.

The total image size after a build on clean repo is
```
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
spades-server       latest              e22412d8d7b7        2 minutes ago       190 MB
```

Some notes:
1. Statically linked libstdc++ so that there's no overhead of extra alpine packages is kept. Maybe there is no so much overhead in terms of size, but semantically feels more right to not depend on extra dynamic libraries in runtime apart from python modules.
2. Adding to `CREDITS` will invalidate pyspades build cache on next docker image build. Not a big deal I guess because not happens often and people will invalidate their cache by other means all the time.
3. `gcc` and `musl-dev` are downloaded twice during docker image build, which increases the build time but generates leaner docker layers (which is better for production set ups and generally pushing to docker hub).

Usage:
1. Make changes of your wish to config and user scripts. (or don't do them at all, the server runs from fresh build just fine)
2. Bake it into image: `docker build -t spades-server .` 
3. Run an instance: `docker run -dP --name spades-server-1 spades-server`
4. Find the mapped ports of the instance: `docker port spades-server-1`
5. Connect to `aos://127.0.0.1:PORT` and enjoy

Future notes:
- Development version which will have fatter layers but less build time and build overhead. Probably this is the job for debian because generally it's a more dev "friendly" distro.
- Make an official image and push it to the [hub.docker.com](http://hub.docker.com)?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -94,6 +94,22 @@ env/
 */*/*.swp
 */*/*/*.swp
 
+# pyspades specific
+pyspades/bytes.cpp
+pyspades/common.cpp
+pyspades/contained.cpp
+pyspades/loaders.cpp
+pyspades/mapmaker.cpp
+pyspades/packet.cpp
+pyspades/vxl.cpp
+pyspades/world.cpp
+enet/enet.c
+
+# native binaries directories
+build/
+enet/build/
+
+
 # Misc
 .github
 .gitattributes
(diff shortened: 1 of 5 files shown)


feikname:
I think we should add an official image as soon as v0.1 is done. I will add it to the roadmap page on wiki.

I am merging these without testing, but If anything happens, we can just revert the commits or fix the bugs :)

Thanks for your contribution!

PR #11, opened 2016-12-30
port config refactor from pysnip

swalladge:
See original explanation at https://github.com/NateShoffner/PySnip/pull/20

:)

----

EDIT from feikname:

*(original text)*

The changes here mostly deal with restructuring how/where confile files, scripts, maps, etc. are loaded from.

 * By default, everything is expected to be under ~/.pysnip (the location can be changed through program arguments)
 * Everything in the configs directory follows a neat structure (as detailed in configs/README.md)
 * I've added a requirements.txt which can be used for installing pip dependencies easily - something could be added to the readme about that (more useful after adding a working setup.py - as in #16 which may need modification after these refactorings)
  * A todo: information about config structure should be added to the readme when finalized how it all should work. A script could also be made to copy example config to ~/.pysnip for getting star


swalladge:
A troubleshooting tip for the wiki if/when this gets merged:

---

#### GeoLiteCity.dat error

If you get the error: `('from' command disabled - missing data/GeoLiteCity.dat)` in the server log, you will need to
download a `GeoLiteCity.dat` file into the data directory in your config directory.

A script is included to automate this - run the following:

```
cd ~/.pysnip/
/path/to/pysnip/feature_server/update_geoip.py
```

swalladge:
Also I should mention that for anyone reviewing this, most of the changes are simply moving existing scripts and other config files to the `config` subdirectory. The actual code changes are in `run.py` and other files under `feature_server`.

feikname:
Going to test the changes right now

feature_server/run.py
```diff
@@ -30,15 +30,17 @@
...
-arg_parser.add_argument("-j","--json-parameters",
-        help="Add extra json parameters, overwriting that in config file.")
-arg_parser.add_argument("-r","--resource-dir", default=".",
-        help="The directory which contains maps,scripts,etc (in correctly named subdirs). - default is in directory of run.py, in feature_server.")
+arg_parser.add_argument("-c","--config-file", default="config.json", 
+        help="specify alternate config file (relative to config dir if relative path)")
+arg_parser.add_argument("-j","--json-parameters", 
+        help="add extra json parameters, overwriting that in config file")
+arg_parser.add_argument("-d","--config-dir", default=os.path.join(os.path.expanduser("~"), ".pysnip"),
+        help="he directory which contains maps,scripts,etc (in correctly named subdirs) - default is ~/.pysnip/")
```
feikname: Typo here

feikname:
```text
Error reading config from /home/flavio/.pysnip/config.json: [Errno 2] No such file or directory: '/home/flavio/.pysnip/config.json'
```

The default config file is not being created anymore.

feature_server/map.py
```diff
@@ -92,9 +89,9 @@ def apply_script(self, protocol, connection, config):
             protocol, connection = self.script(protocol, connection, config)
         return protocol, connection
 
-    def load_vxl(self, rot_info, load_dir):
+    def load_vxl(self, rot_info):
         try:
-            fp = open(rot_info.get_map_filename(load_dir), 'rb')
+            fp = open(rot_info.get_map_filename(self.load_dir), 'rb')
```
noway421: `self.load_dir` is not set to an instance anyware.. Were those 2 line changes unnecessary? Rolling back those 2 lines fixed map loading.

feature_server/run.py
```diff
@@ -62,17 +70,17 @@ def choose_path(base,top):
...
+    with open(config_file, 'rb') as f:
+        config = json.load(f)
+        cfg.config = config
+except IOError as e:
+    print("Error reading config from {}: ".format(config_file) + str(e))
+    sys.exit(1)
+except ValueError as e:
+    print("Error in config file {}: ".format(config_file) + str(e))
+    sys.exit(1)
+
```
noway421: config.default.txt is not copied on first launch. but not a big deal, we can merge it with out it

feature_server/map.py
swalladge: ah i think load_information was supposed to fix that - hang on i'll add it

feikname:
@noway421 was it tested in-game already?

noway421:
yes

PR #14, opened 2017-01-01
Pysnip as a module

noway421:
build instructions:
1. install virtualenv just as usual:
```
virtualenv -p python2 venv
. venv/bin/activate
pip install -r requirements.txt
```
2. build&install pysnip
```
pushd enet && python setup.py build_ext --inplace && popd && python setup.py build && python setup.py install
```
3. try it out:
```
pysnip -d  /path/to/config/dir
```

notes: 
1. potentially make sys.path handling more tidy? right now run.py puts absolute path of the egg in sys.path
2. use find_packages?
3. naming is a big question, how to name the commands, the packages better etc, etc.
4. `web` directory is not customisable in this build. should it be? don't know, i reckon probably not, not many people customize it although some do a lot of customization in there.


feikname:
Could you also add the instructions mentioned above to the README.md?

Also, no maps nor scripts are loading for me, do they need to be in the config directory?

----

1. absolute path is fine by now
2. sure, why not?
3. aren't the commands already named? :confused: We can discuss this on Slack
4. i think we should leave `web` personalization to a later time

.gitignore
```diff
@@ -23,6 +23,8 @@ build/
 enet/build/
 venv/
 env/
+dist/
+pysnip.egg-info/
```
swalladge: might as well be `*.egg-info/` to avoid problems if name changes

feature_server/run.py
```diff
@@ -57,6 +57,10 @@ def choose_path(base,top):
 
 # add it to the path so we can import scripts
 sys.path.append(config_dir)
+# add our package to path too so scripts can import `feauture_server/`
```
swalladge: is this necessary? I would have thought if installed as a python package (or running in a virtualenv), `feature_server` should be available to import regardless.

setup.py
```diff
@@ -0,0 +1,53 @@
...
+
+setup(
+    name = 'pysnip',
+    packages = ['pysnip', 'pysnip.web', 'pyspades', 'pyspades.enet'],
+    version = '0.0.0',
+    description = 'Open-source server implementation for Ace of Spades',
+    author = 'Matpow2, Stackoverflow',
+    author_email = 'nate.shoffner@gmail.com',
+    url = 'https://github.com/NateShoffner/PySnip',
+    download_url = 'https://github.com/NateShoffner/PySnip/archive/master.tar.gz',
```
swalladge: might want to update this info for spades-server - including the `name` to `spades-server`. Obviously keep info about original authors and copyright, but probably should go in the readme.

setup.py
```diff
@@ -0,0 +1,53 @@
...
+	setup_requires = ['cython'],
+	install_requires = ['twisted'],
+	extras_require = {
+		'from': ['pygeoip'],
+		'statusserver': ['jinja2', 'pillow'],
+		'ssh': ['pycrypto', 'pyasn1']
+	},
+    entry_points = {
+        'console_scripts': [
+            'pysnip=pysnip.__main__:main'
```
swalladge: Also update all pysnip references to pyspades. If you want to make it easy to revert in case of backporting to pysnip, add a global variable like `PKGNAME='spades-server'` and use that throughout the file.

swalladge: Also could be simpler if this was changed to `'spades-server=feature_server.run'`. Don't need an extra file then.

feature_server/run.py
noway421: Yes, but it would be accessible as `import pysnip; print pysnip.commands` (pysnip=feature_server, see line https://github.com/feikname/spades-server/pull/14/files#diff-2eeaed663bd0d25b7e608891384b7298R49). The way I'm doing it there allows scripts to access it like this: `import commands`

setup.py
noway421: Yep, that's the question of naming. While download_url, author, author_email, description should be tailored for us, I'm not sure about name.

Leave as `pysnip`? Change it to `spades-server`? In the second case do we want to change it eveywhere or just for package name... Also, `spades-server` is not an advisable name for a package name because of the following: http://stackoverflow.com/questions/761519/is-it-ok-to-use-dashes-in-python-files-when-trying-to-import-them

setup.py
noway421: Good idea about PKGNAME 👍 

I couldn't reference feature_server.run (or actually in this case it would be pysnip.run because pysnip=feature_server) because run.py doesn't have main and it just fires things away on import time. A better way might be to add `def main(): ... ; __name__ == __main__: main()` into run.py..

.gitignore
noway421: 👍

feature_server/run.py
swalladge: better make sure none of the files then conflict with other packages/modules then. I'm especially worried about `map.py`, `cfg.py`, `run.py`....
I think it could be safer to keep the namespacing.

You can always import like this still:
```
from pysnip import commands, console, irc
print commands
```

noway421: It is true that this importing is a better way and probably something we would advice users/developers to move towards. But for legacy `import commands` should work too.

I'll try to scope the importing to userscripts only (was my intention from the start, but looking at the current implementation it's not here). In userscripts, it is expected that `import map` will import map.py from pyspades and not some https://pypi.python.org/pypi/map

PR #29, opened 2017-01-13
Fix Typo in Dockerfile

NotAFile:


noway421:
That's not really a typo, but the path of the volume inside docker container. You can see that `VOLUME /config` is there in the docker file

At the same time, I think it would be a good idea to move using `config/` name for the config directory instead of `configs/`? Or is it? 

Probably better to use `configs/` still cause there can be multiple config files in there. But also there are resources there as well...

NotAFile:
Oh, yes. Docker gave me an error that it could not find `config/config.json`. I shelled in and saw a directory named `configs` and assumed it was a typo. Probably should have actually tested it beforehand.

The actual /config/directory is completely empty for me, which should probably be corrected in either the instructions or the code. Anyway, this should probably be transferred to an issue instead.

feikname:
@NotAFile The config file was renamed from config.txt to config.json, and the old code is still trying to copy the old one to the newly created directory, which results in an error. I am probably going to submit a fix very soon.

noway421:
Oups, sorry, not updated the docs yet! Yes, you need to specify volume for docker container with the `-v` option like this: `-v ~+/configs/:/config`

noway421:
I'm leaving `configs/` directory for config files. It makes sense for them to be there.

Issue #31, opened 2017-01-13
prebuild.sh to python


noway421:
prebuild.sh not copied to source code dist. we can copy the shell script with package_data but might as well rewrite it to python

PR #35, opened 2017-01-15
prebuild.sh to python

noway421:
Instead of relying on a bash script, rewrote it to python. 

This is a prep for release.


feikname:
LGTM

setup.py
```diff
@@ -38,12 +41,51 @@
...
+        os.remove("pyenet/enet.so")
+    if os.path.isdir("pyenet/enet"):
+        shutil.rmtree("pyenet/enet")
+
+    shutil.copyfile("pyenet/enet.pyx", "pyenet/enet-pyspades.pyx")
+    subprocess.Popen(['patch', '-p1', 'pyenet/enet-pyspades.pyx', 'pyspades-pyenet.patch']).communicate()
+
+    if not os.path.isfile(enet_file):
+        print("Downloading enet")
+        urllib_request.urlretrieve(enet_url, enet_file)
```
NotAFile: also, we should note that this means that there is no longer a progress bar for the download. I don't think it's a huge problem.

NotAFile: Turns out it's pretty simple to add a progress bar: https://stackoverflow.com/questions/13881092/download-progressbar-for-python-3

.gitignore
```diff
@@ -53,3 +54,6 @@ enet/pyenet/enet-pyspades.pyx
 enet/pyenet/enet/
 enet/pyenet/build/
 enet/enet-*.tar.gz
+
+#misc
+.DS_Store
```
feikname: This line should end with a newline character.

setup.py
noway421: it complicates the script, I would prefer to keep it as lean as possible. "Downloading enet" is quite descriptive and doesn't take too long for user to keep wondering what's going on. The archive is relatively small

feikname:
@NotAFile I converted your requested change to an [issue](https://github.com/piqueserver/piqueserver/issues/36)

feikname:
Just a note that wasn't included here, this fixes #31.

PR #38, opened 2017-01-16
[WIP] pip fixes

noway421:
The main issue is that cython is being installed, but not being picked up by setup.py because the setup.py script is still running... would be nice if setup.py would just restart after `egg_info` command.

Tried to defer the import of cython, but then can't find a way to make cythonize work

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -8,10 +8,10 @@ recursive-include pyspades/ *
 
 global-exclude *.pyc
 global-exclude *.pyo
-global-exclude *.pyd
+# global-exclude *.pyd
 
 # C extensions
-global-exclude *.so
+# global-exclude *.so
 
 # C++ extension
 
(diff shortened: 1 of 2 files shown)


noway421:
this:https://github.com/h5py/h5py/blob/master/setup_build.py#L195

noway421:
fixed cython, but now enet.so is not getting copied over

swalladge:
I believe commit https://github.com/piqueserver/piqueserver/pull/38/commits/b17eb5e63fa2b4e8a4484fbc265b8b31ff9c9200 has fixed enet.so not being copied. The package_data entry didn't seem to do anything, so removed that for now. Now to work out if any other files in the `pyspades/enet/` directory needs to be ignored.

noway421:
Do you need some more time or are we on the straight road to PyPi?

PR #39, opened 2017-01-18
Pillow fixes

swalladge:
- allow using Pillow 4.0 in dependencies
- fix method call (fromstring hasn't been working since Pillow 3.0)

### DIFF
--- a/feature_server/statusserver.py
+++ b/feature_server/statusserver.py
@@ -116,7 +116,7 @@ class StatusServerFactory(object):
         self.last_map_name != self.protocol.map_info.name or
         current_time - self.last_overview > OVERVIEW_UPDATE_INTERVAL):
             overview = self.protocol.map.get_overview(rgba = True)
-            image = Image.fromstring('RGBA', (512, 512), overview)
+            image = Image.frombytes('RGBA', (512, 512), overview)
             data = StringIO()
             image.save(data, 'png')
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 Cython>=0.25.2,<0.26
 Twisted>=16.6.0,<16.7
 Jinja2>=2.8,<2.9
-Pillow>=3.4.2,<3.5
+Pillow>=3.4.2,<=4.0
```
noway421: should be starting from 4.0, shouldn't it? You are using new method, so should be new version range

feature_server/statusserver.py
```diff
@@ -116,7 +116,7 @@ def get_overview(self):
         self.last_map_name != self.protocol.map_info.name or
         current_time - self.last_overview > OVERVIEW_UPDATE_INTERVAL):
             overview = self.protocol.map.get_overview(rgba = True)
-            image = Image.fromstring('RGBA', (512, 512), overview)
+            image = Image.frombytes('RGBA', (512, 512), overview)
```
noway421: 👍

requirements.txt
swalladge: frombytes and the removal of fromstring has been in Pillow since 3.0

noway421: 👍

PR #42, opened 2017-01-18
initial relaxing of dependency version restrictions

swalladge:
This relaxes most of the dependency restrictions to major version checks only. A quick search around didn't raise any flags for breaking changes within the new ranges, and it runs fine with the official Arch repos versions of packages.


noway421:
LGTM 👍

PR #48, opened 2017-01-23
Pyenet update

noway421:
Updated to latest pyenet `master`. Tested with gamedig, `query.py` is working.

### DIFF
--- a/enet/pyenet/.gitignore
+++ b/enet/pyenet/.gitignore
@@ -1,0 +1,95 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+env/
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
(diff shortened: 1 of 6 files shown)


NotAFile:
I'm not sure how useful it is to set this up considering our goal is to get pyenet onto pypi asap, but this kind of stuff is what git submodules were made for. Doing it this way lgtm for now though.

noway421:
I agree with you completely re: both points. Git submodules are a good fit here, yes, but I personally find them very brittle and not easy to use (eg need more than simple clone to fetch the whole repo, problems with versioning). 

The goal is to get pyenet in pypi, yes, and I view this PR as a way to track that progress. It is an iterative step. We need the .patch file to track what changes are left to be merged, so might as well commit that to our `master`.

swalladge:
I don't know how cython works so maybe I'm not the best reviewer here. :P Either way, it looks great and is a step closer to being able to use it directly from upstream, so nice work! :)

feikname:
@noway421 is this ready to merge?

noway421:
👍

Issue #44, opened 2017-01-19
update update_geoip.py script


swalladge:
The `update_geoip.py` script is designed to automatically download the geoip database into the `data` directory, but it needs to be updated to use the new config directory.

A good idea would be to add another commandline flag or script to the package to be able to call it from `feature_server/run.py`.

swalladge:
Fixed in #49

PR #49, opened 2017-01-23
Refactor run2

swalladge:
note: the first branch I had merged in from the pyenet-updates branch as well for some reason - this one is based purely off master.

Anyway, this refactors out the cli args handling to
`feature_server/__main__.py`, making `run.py` neater and helping to separate
the actual server part from args handling and other startup tasks not related
to running the server.

It also moves and modifies the `update_geoip` script to work with the new
config directory locations, along with the new `--update-geoip` option which
calls the script.

Not exactly related to the refactor, but I also changed how json parameters
passed to the `-j` option are loaded: now using the more secure `json.loads`
rather than `eval`.


swalladge:
note: this fixes issue #44

NotAFile:
I am not a big fan of moving the argument parsing logic to __main__.py. As I understand it, that is not what __main__.py files are made for, and it makes running the server in it's folder directly more confusing.

noway421:
As from discussion from slack:

I believe the best way is when it is working without __main__.py, but __main__.py is still there for the support of `python2 -m piqueserver.feature_server`-style execution (edited)

noway [1:18 PM] 
I think there is no need to split up run.py, but instead refactor it so that there are `def main():` for running your argument parser whole-thing and `def server():` (think of it as server_main) which runs just the server part

and then in __main__.py we only have 3 lines:

 ```import run
if __name__ == '__main__':
  run.main()
```

swalladge:
mumble mumble merge conflicts... just rebased this pull request on master, so merging should be fine. Various parts of `run.py` and `__main__.py` were conflicting since some strings were edited on master while I was juggling things between files.

feikname:
Feel free to merge this. I don't have the sufficient Python knowledge to judge this pull request yet :P

PR #57, opened 2017-01-27
Tidy server.py + better script loading

swalladge:
- tidies up top level code in `server.py` so that importing it doesn't have side effects (except still for epollreactor installing for twisted - maybe something who knows twisted can fix that sometime)
- cleaner importing scripts using `imp` - no manipulating `sys.path` anymore!
- the original changes to `sys.path` was allowing the scripts to still work with importing files directly from `feature_server` like `import commands` - now you need to `from piqueserver import commands` or `import piqueserver.commands` from scripts.

### DIFF
--- a/feature_server/config/scripts/afk.py
+++ b/feature_server/config/scripts/afk.py
@@ -8,7 +8,7 @@ from math import ceil
 from operator import attrgetter
 from twisted.internet import reactor
 from pyspades.common import prettify_timespan
-from commands import name, get_player, add, admin
+from piqueserver.commands import name, get_player, add, admin
 
 S_AFK_CHECK = '{player} has been inactive for {time}'
 S_NO_PLAYERS_INACTIVE = 'No players or connections inactive for {time}'
--- a/feature_server/server.py
+++ b/feature_server/server.py
@@ -21,6 +21,7 @@ pyspades - default/featured server
 
 import sys
 import os
+import imp
 import json
 import itertools
 import random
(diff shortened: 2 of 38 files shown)


feature_server/config/scripts/afk.py
```diff
@@ -8,7 +8,7 @@
 from operator import attrgetter
 from twisted.internet import reactor
 from pyspades.common import prettify_timespan
-from commands import name, get_player, add, admin
+from piqueserver.commands import name, get_player, add, admin
```
noway421: I do agree that it is The Right Way To Move Forward.

but are we keeping the backwards compatibility? I think it is important for people to be able just plug in the scripts from their existing pyspades/pysnip installation without tweaking imports

feature_server/server.py
```diff
@@ -1048,10 +1020,13 @@ def run():
...
         script_names.append(game_mode)
 
+    script_dir = os.path.join(cfg.config_dir, 'scripts/')
     for script in script_names[:]:
         try:
-            module = __import__('scripts.%s' % script, globals(), locals(),
-                [script])
+            # NOTE: this finds and loads scripts directly from the script dir
+            # no need for messing with sys.path
+            f, filename, desc = imp.find_module(script, [script_dir])
```
noway421: wicked 👍

swalladge: Yeah it's pretty neat. 😄  Discovered at https://pymotw.com/2/imp/

feature_server/config/scripts/afk.py
swalladge: ok, re-added to keep backwards compatibility.

PR #58, opened 2017-01-27
More refactoring

feikname:
The intention of this pull request is to make piqueserver code PEP8 compliant as much as possible.


feikname:
I also changed the `feature_server` headers because I think they are more informative this way, but I can revert them if anyone wishes to.

NotAFile:
I can't possibly look at all of this stuff but it seems to be good. This will create a ton of merge errors everywhere, but it needs to be done at some point.

noway421:
👍

noway421:
Too soon. Have to revert it. We will get it back.

PR #61, opened 2017-01-28
Infogulch elaboration

noway:
That's all of it pretty much. 

A lot of whitespace are old white space. We'll fix them up as extra white space fixing commits in master.

## How to view this PR

Don't be scared by +1,359 -1,164 most of them are whitespace noise we would fix later on.

View the diff like this: https://github.com/piqueserver/piqueserver/pull/61/files?w=1 (or like this https://github.com/piqueserver/piqueserver/pull/61/files?diff=split&w=1)

Boils down to +332 −137 coherent changes

## Yay

review and comment pls 😊


feikname:
@noway421 could you please write a summary the changes?

noway421:
Erh ah, can't really track down every change right now (but will try, by looking at the commit messages), but they are changes that infogulch and co did over time to improve pyspades which was running on their servers.

feikname:
Yeah, I realized right now hey're quite descriptive. There is no need for a summary.

feature_server/commands.py
```diff
@@ -197,10 +213,16 @@ def say(connection, *arg):
     connection.protocol.send_chat(value)
     connection.protocol.irc_say(value)
 
-@admin
-def kill(connection, value):
+add_rights('kill', 'admin')
```
swalladge: Should this be a decorator rather than a function call?

feature_server/map.py
```diff
@@ -110,10 +111,10 @@ def __init__(self, name = "pyspades"):
         self.name = name
 
     def get_seed(self):
-        if self.seed is None:
-            random.seed()
-            self.seed = random.randint(0, math.pow(2, 31))
+        if self.seed is not None:
```
swalladge: Don't know what has happened here... definitely not right.

pyspades/server.py
```diff
@@ -1007,9 +1015,9 @@ def grenade_exploded(self, grenade):
         z = position.z
         if x < 0 or x > 512 or y < 0 or y > 512 or z < 0 or z > 63:
             return
-        x = int(x)
-        y = int(y)
-        z = int(z)
+        x = int(math.floor(x))
```
swalladge: somewhat redundant

feature_server/commands.py
```diff
@@ -197,10 +213,16 @@ def say(connection, *arg):
...
+def kill(connection, value = None):
+    if value is None:
+        player = connection
+    else:
+        if not connection.rights.kill:
+            return "You can't use this command"
     player = get_player(connection.protocol, value, False)
     player.kill()
+    if connection is not player:
```
swalladge: Indentation errors here. I'm guessing the next 2 lines should be indented and thus controlled by this `if`, but I'm not entirely sure what logic is wanted here.

feature_server/map.py
swalladge: Fixed in d86e179

pyspades/server.py
noway421: Maybe NaN safeguard? Just guessing

swalladge:
Whitespace issues fixed now; viewing the diff should be much nicer.

pyspades/server.py
feikname: `math.floor()` returns a float value instead of an integer one in Python2

swalladge: Still wouldn't help - `int(math.floor(float('nan')))` still crashes the same as `int(float('nan'))`.

noway421: probably the good old arithmetic parity problem of int cast:
```
>>> int(-0.5)
0
>>> import math
>>> math.floor(-0.5)
-1.0
```

it's needed in those vector calculations two avoid off by one block error I guess

swalladge: ah that could be it, fair enough

feature_server/commands.py
noway421: lesson learnt: 'rebase ignore space change' and python are not compatible. Yes, there are quite a few places where this holds true and their are screwed in terms of indentation.


Those are the places where there are changes I believe. There will be no places where it will have silently screwed, there gotta be change there (because we are comparing against our prestine `master`)

feature_server/commands.py
noway421: Previous version: only admins can use
Proposed version: regular players can commit suicide and admin can kill whoever they want

swalladge:
Looks to me like it's just that weird indenting in `commands.py`, and then it all should be fine. :)

feikname:
Seems like I lost the review comments...

pyspades/vxl.pyx
```diff
@@ -118,16 +118,17 @@ cdef class VXLData:
 
     def destroy_point(self, int x, int y, int z):
         if not self.get_solid(x, y, z) or z >= 62:
-            return False
+            return 0
```
feikname: **TODO:** Check the implications of this change

feature_server/server.py
```diff
@@ -324,13 +324,17 @@ def on_team_join(self, team):
...
+            if not team.spectator and not team.other.locked:
+                return team.other
             return False
         balanced_teams = self.protocol.balanced_teams
         if balanced_teams and not team.spectator:
             other_team = team.other
             if other_team.count() < team.count() + 1 - balanced_teams:
-                self.send_chat('Team is full')
-                return False
+                if other_team.locked:
```
feikname: **TODO:** Double-check this logic later

feature_server/server.py
```diff
@@ -324,13 +324,17 @@ def on_team_join(self, team):
                     return False
         if team.locked:
             self.send_chat('Team is locked')
+            if not team.spectator and not team.other.locked:
```
feikname: **TODO:** Double-check this logic later

feature_server/config/scripts/grownade.py
```diff
@@ -97,6 +91,23 @@ def model_grenades(connection, expression = None):
...
+        raise ValueError()
+    player = connection
+
+    result = None
+    if expression:
+        try:
+            models, loaded, failed = load_models(expression)
+        except ModelLoadFailure as err:
+            result = str(err)
+        else:
```
feikname: Something seems wrong here (indentation errors too?)

feature_server/config/scripts/arena.py
```diff
@@ -343,6 +387,13 @@ def arena_win(self, team, killer = None):
             self.send_chat(team.name + ' team wins the round!')
             self.begin_arena_countdown()
 
+        def arena_reset_fog_color(self):
+            if self.arena_old_fog_color is not None:
+                # Shitty fix for disco on game end
```
feikname: This comment ^

feature_server/config/scripts/arena.py
```diff
@@ -36,29 +44,40 @@
...
+import random
+import math
 
 # If ALWAYS_ENABLED is False, then the 'arena' key must be set to True in
 # the 'extensions' dictionary in the map metadata
 ALWAYS_ENABLED = True
 
 # How long should be spent between rounds in arena (seconds)
-SPAWN_ZONE_TIME = 20.0
+SPAWN_ZONE_TIME = 15.0
```
feikname: For the sake information, if anyone is against chaning these default values, please speak up. There are more changed below, but I'm only adding a comment to this one.

feature_server/commands.py
```diff
@@ -979,42 +1014,30 @@ def handle_command(connection, command, parameters):
...
+        print t
+        return 'Command failed'
     except InvalidPlayer:
         return 'No such player'
     except InvalidTeam:
         return 'Invalid team specifier'
     except ValueError:
         return 'Invalid parameters'
 
-def debug_handle_command(connection, command, parameters):
```
feikname: Will the removal of this function affect anything?

feature_server/commands.py
```diff
@@ -979,42 +1014,30 @@ def handle_command(connection, command, parameters):
         command_func = commands[command]
     except KeyError:
         return # 'Invalid command'
+    aspec = command_func.argspec
```
feikname: Can we rename the local variable here from `aspec` instead of `argspec`? I'm almost certain it was a typo.

feature_server/commands.py
```diff
@@ -524,6 +560,11 @@ def god(connection, value = None):
         connection.god_build = connection.god
     else:
         connection.god_build = False
+    return 'You have entered god mode.'
+
+@admin
+def god(connection, value = None):
+    godsilent(connection, value)
```
feikname: I don't undertand why is `godsilent` called from within `god`, can anhone please explain?

feature_server/commands.py
```diff
@@ -453,7 +480,8 @@ def teleport(connection, player1, player2 = None, silent = False):
         silent = silent or player.invisible
         message = '%s ' + ('silently ' if silent else '') + 'teleported to %s'
         message = message % (player.name, target.name)
-    player.set_location(target.get_location())
+    x, y, z = target.get_location()
+    player.set_location(((x-0.5), (y-0.5), (z+0.5)))
```
feikname: The reason behind this should be analyzed (?)

pyspades/vxl.pyx
```diff
@@ -118,16 +118,17 @@ cdef class VXLData:
...
         start = time.time()
         for node_x, node_y, node_z in self.get_neighbors(x, y, z):
             if node_z < 62:
-                self.check_node(node_x, node_y, node_z, True)
+                count += self.check_node(node_x, node_y, node_z, True)
         taken = time.time() - start
         if taken > 0.1:
             print 'destroying block at', x, y, z, 'took:', taken
-        return True
+        return count
```
feikname: **TODO:** Check the implications of this change

pyspades/vxl_c.cpp
```diff
@@ -149,7 +149,7 @@ int check_node(int x, int y, int z, MapData * map, int destroy)
         z = current_node->z;
         if (z >= 62) {
             marked.clear();
-            return 1;
+            return 0;
```
feikname: **TODO:** Check the implications of this change

pyspades/vxl_c.cpp
```diff
@@ -180,17 +180,18 @@ int check_node(int x, int y, int z, MapData * map, int destroy)
         }
     }
     
+    int ret = (int)marked.size();
     marked.clear();
-    return 0;
+    return ret;
```
feikname: **TODO:** Check the implications of this change

feature_server/config/scripts/arena.py
swalladge: This is fairly low priority to worry about - scripts are supposed to be easily modifiable for the server administrator (hence these settings being at the top as constants in the first place). Looks like the new values are reasonable defaults.

pyspades/vxl.pyx
swalladge: I believe that this is to support the `total_blocks_removed` attribute. The function is currently used in `scripts/platform.py` in `if` statements, which is fine because 0 is falsey, and non-zero is truthy. Where it is used in `server.py` has been updated by infogulch to suit.

pyspades/vxl.pyx
swalladge: Related to previous comment on change just above this.

feature_server/commands.py
swalladge: Fixed

feature_server/config/scripts/grownade.py
swalladge: Yep definitely - I think I've fixed it, but I guess the only way to check for logic problems would be to thoroughly test the script

feature_server/commands.py
swalladge: Renamed.

feature_server/commands.py
swalladge: The only place it seems to have been used is below it, and that's commented out. My guess it was once used a long time ago, but now it is well and truly redundant. (the `handle_command` function does way more than this function did anyway)

feature_server/commands.py
swalladge: `god` is a wrapper for the `godsilent` command that sends the message in chat rather than toggling god mode silently.

pyspades/vxl_c.cpp
swalladge: This is related to the `check_point` function returning a count instead of bool. It's used in the `destroy_point` function where its return value is used to increment the count of blocks destroyed. Looks fine to me.

pyspades/vxl_c.cpp
swalladge: Related to comment above re count.

feature_server/commands.py
swalladge: This looks slightly worrying - what if the target player was standing right next to a wall and (x-0.5) gets the teleporting player stuck in the wall? Perhaps the change was due to a player somehow getting stuck in the other player or in the ground - I think we need to do some testing before merging this change.

feature_server/commands.py
noway421: Indeed!

feature_server/config/scripts/arena.py
noway421: Agree, my thoughts exactly

feature_server/config/scripts/arena.py
```diff
@@ -260,11 +279,29 @@ def on_team_join(self, team):
...
+                    if min_distance is None or distance < min_distance:
+                        min_distance = distance
+                if min_distance > self.protocol.arena_max_spawn_distance:
+                    self.set_location(random.choice(self.team.arena_spawns))
+                    self.refill()
+            return connection.on_position_update(self)
+
         def get_respawn_time(self):
+            if self.protocol.arena_enabled:
             if self.protocol.arena_running:
```
noway421: bad indentation should be this:
```python
         def get_respawn_time(self):
            if self.protocol.arena_enabled:
                if self.protocol.arena_running:
                    return -1
                else:
                    return 1
            return connection.get_respawn_time(self);
```

(see https://github.com/NateShoffner/PySnip/pull/6/files#diff-13ae464b925f0aca4965862189c7a1c7R299)

feature_server/config/scripts/arena.py
```diff
@@ -332,8 +371,13 @@ def arena_time_limit(self):
         def arena_win(self, team, killer = None):
             if not self.arena_running:
                 return
+            if self.arena_old_fog_color is None and TEAM_COLOR_TIME > 0:
+                self.arena_old_fog_color = self.fog_color
+                    self.set_fog_color(team.color)
```
noway421: bad indent see https://github.com/NateShoffner/PySnip/pull/6/files#diff-13ae464b925f0aca4965862189c7a1c7R375

feature_server/config/scripts/arena.py
```diff
@@ -332,8 +371,13 @@ def arena_time_limit(self):
...
+                    reactor.callLater(TEAM_COLOR_TIME, self.arena_reset_fog_color)
             if killer is None or killer.team is not team:
                 for player in team.get_players():
+                    if not player.world_object.dead:
                     killer = player
```
noway421: bad indent see https://github.com/NateShoffner/PySnip/pull/6/files#diff-13ae464b925f0aca4965862189c7a1c7R381

feature_server/server.py
noway: seems fine. consider this: you selecting blue but it is locked. it will throw you to green

feature_server/server.py
noway: Seems fine. If blue is full it will transfer to green unless green is locked

feature_server/config/scripts/grownade.py
noway: fixed by swalladge

feature_server/map.py
```diff
@@ -111,10 +111,10 @@ def __init__(self, name = "pyspades"):
         self.name = name
 
     def get_seed(self):
-        if self.seed is None:
-            random.seed()
-            self.seed = random.randint(0, math.pow(2, 31))
-        return self.seed
+        if self.seed is not None:
+            return self.seed
+        random.seed()
```
NotAFile: this version does not actually change self.seed, is this correct?

feature_server/commands.py
noway: yeah fair enough, but it is admin teleport, not so bad. Yeah, you can get stuck, but it is z+0.5 so you might pop out. certainly gives better admin experience - that's for sure, you are not spawning inside the player which is weird.

feature_server/map.py
noway: Thanks for spotting, but it is expected behaviour:

https://github.com/NateShoffner/PySnip/pull/6/commits/9ede53f82d52b89416841f48c02b729fd0249895
https://code.google.com/archive/p/pyspades/issues/282

feature_server/commands.py
NotAFile: maybe `player.get_location()` gives the coordinates in block space, not player space?

feature_server/commands.py
NotAFile: It's done the other way around for e.g. `/gotos`, so this might want to be adjusted.

On the other hand, the question is if this command should really exist, since I presume the reason the chat message was there originally was to prevent admin cheating.

feature_server/commands.py
noway: Tested right now, I'm being teleported exactly into the player, not even any bounce can be observed at all. So this code must be correct.

feature_server/config/scripts/arena.py
NotAFile: Ideally the scripts would use the config system too, I think.

feature_server/commands.py
NotAFile: I just tested the older version, and I get stuck in the bottom of the block

feature_server/map.py
NotAFile: soo... why not fix it?

noway:
#huge

PR #67, opened 2017-02-06
Updated nosharp.py

kmsiapps:
- Added options to turn on/off
- Now puts player on spectator team instead of kicking

### DIFF
--- a/feature_server/config/scripts/nosharp.py
+++ b/feature_server/config/scripts/nosharp.py
@@ -1,25 +1,39 @@
 """
 nosharp.py - kicks player whose name starts with # and also kicks player with no name
 by kmsi(kmsiapps@gmail.com)
-version 2(2017.02.06)
- - Added options to turn on/off
+version 3(2017.02.06)
+ - Added options
  - Now puts player on spectator team instead of kicking
+ - Auto-disconnects after specific time(default : 10 seconds)
 """
+from twisted.internet import reactor
+
 def apply_script(protocol,connection,config):
     class noSharpConnection(connection):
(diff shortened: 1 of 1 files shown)


swalladge:
The only problem I can see is that players with invalid names can sit in the spectators filling up the connection (are players' slots filled by spectators?) and still can't be kicked

noway:
Yep. Autokick in 10 secs would be nice

NotAFile:
maybe an autokick after, say, 15 seconds would be a good idea

kmsiapps:
Added autokick.

feature_server/config/scripts/nosharp.py
```diff
@@ -1,25 +1,39 @@
...
-                    
-            return connection.on_spawn(self, pos)
+                    self.send_chat('%% Your name starts with #. Will be kicked soon.')
+                    self.autokick_call = reactor.callLater(autokick_duration, self.autokick)
+            else:        
+                return connection.on_spawn(self, pos)
+                
+        def autokick(self):
+            self.autokick_call = None
+            self.kick('Invalid name')
```
NotAFile: have you checked if this reason works in both OS and AoS? Last I heard, arbitrary reasons were not allowed.

yvt: Arbitrary kick reasons aren't supported by neither of AoS and OS. I reckon it's only displayed on other players' chat log window and the server log.

kmsiapps: @NotAFile It is shown in-game to other players.

PR #69, opened 2017-02-07
Move bugfix.py to core server

swalladge:
Implement `bugfix.py` script fixes in the core server (take 2 due to epic git fail in #68 ).


swalladge:
Strangely this seems to disable line building altogether... :\

feature_server/server.py
```diff
@@ -205,7 +207,16 @@ def on_block_build_attempt(self, x, y, z):
         return self._can_build()
 
     def on_line_build_attempt(self, points):
-        return self._can_build()
+        if not self._can_build():
```
yvt: Surprisingly, `self._can_build()` [returns `None`](https://github.com/piqueserver/piqueserver/blob/master/feature_server/server.py#L198) if building is allowed. This line should be replaced by `if self._can_build() == False:` to get the desired behavior.

swalladge: oooh i see. So `on_line_build_attempt` must onnly ever be checked for false as well... weird

swalladge: Yep, that was it. It works fine now. :)

piqueserver/server.py
```diff
@@ -211,7 +215,16 @@ def on_block_build_attempt(self, x, y, z):
         return self._can_build()
 
     def on_line_build_attempt(self, points):
-        return self._can_build()
+        if self._can_build() == False:
```
noway: looks solid

PR #73, opened 2017-02-14
Add info about AUR package

swalladge:
This modifies the layout of the installation section slightly and adds info about the archlinux aur package available. Not sure if we want to keep adding this info to the readme, or put installation info into the wiki - it's going to get pretty lengthy in the readme once there are more packages available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@ python setup.py install
 
 ### Archlinux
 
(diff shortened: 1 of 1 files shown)


feikname:
The changes were approved by 3 members, so it's ready to merge I guess.

PR #74, opened 2017-02-14
Added scripts

kmsiapps:
- nospadingwalls.py (prevents spading through walls)
- nointelonwalls.py (prevents grabbing intel through walls)


piqueserver/config/scripts/ratio.py
```diff
@@ -5,47 +5,75 @@
...
+        ratio_headshotkills = 0
+        ratio_meleekills = 0
+        ratio_grenadekills = 0
         ratio_deaths = 0
 
         def on_kill(self, killer, type, grenade):
             if killer is not None and self.team is not killer.team:
                 if self != killer:
                     killer.ratio_kills += 1
+                    killer.ratio_headshotkills += type == HEADSHOT_KILL
```
feikname: Can anyone explain the syntax used here? (a link to the docs would also be welcome)

swalladge: It's adding the result of `type == HEADSHOT_KILL` to `killer.ratio_headshotkills`. As `True == 1`, this will increment `ratio_headshotkills` by one if it's a headshot.

feikname: @swalladge I undertand now thanks!

Issue #76, opened 2017-02-25
Imports should be working both ways


noway:
Seems to be a regression on our end. 

>@1AmYF
>Another import thing - i stumbled upon the command /ratio not working, and saw that it seems necessary now to import from `piqueserver.commands` instead of `commands`only. This effects the compatibility of a quite large amount of scripts written for pyspades/pysnip. I noticed that some of the scripts delivered with pique have that import already adjusted, does that mean that it's not considered an issue? :cold_sweat:

PR #79, opened 2017-02-26
An attempt at fixing commands import

noway:
Fixing #76 

View this PR as https://github.com/piqueserver/piqueserver/pull/79/files?w=1

essentially the main problem is that the modules are loaded under different names.

`commands`
`piqueserver.commands`

if that happens, python interpreter treats them as 2 different modules and they end up having 2 different copies in memory.

For example, this is the addresses for the function commands.add:
```
('__name__', 'piqueserver.commands')
('add', <function add at 0x7f496a182aa0>)
('__name__', 'commands')
('add', <function add at 0x7f4969eec398>)
```

This PR fixes them to be the same:
```
('__name__', 'piqueserver.commands')
('add', <function add at 0x7fa770fbdaa0>)
('__name__', 'commands')
('add', <function add at 0x7fa770fbdaa0>)
```

### DIFF
--- a/piqueserver/commands.py
+++ b/piqueserver/commands.py
@@ -17,1120 +17,1124 @@
 
 # pylint: disable=too-many-lines
 
-import os
-import math
-import inspect
-from random import choice
+if __name__ == 'commands':
+    from piqueserver.commands import *
+else:
 
-from twisted.internet import reactor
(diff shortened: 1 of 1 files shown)


noway:
damn pylint

PR #81, opened 2017-03-04
Updates, additions, fixes for gamemodes and scripts

1AmYF:
- Updated votekick.py to newer version including the command /togglevotekick
- Updated arena and babel gamemode (originated from aloha.pk)
- Added support for bots by adding the attribute 'local' to ServerConnection, allows scripts to create and distinguish local connections
- Added optional config parameter 'remove_intel' to tdm, to allow non-ctf-related gameplay
- Added the push gamemode
- Fixed passreload.py (was not working due to refactorings concerning the config)

PR #85, opened 2017-04-19
Small fixes related to PR #81

1AmYF:
2 (non-critical) fixes related to changes made in #81:

- Use actual spawn range for intel (instead of the smaller range for bases)
- Import from piqueserver


noway:
Dunno what's up with the travis, but LGTM

feikname:
> Dunno what's up with the travis, but LGTM

It's `pylint` getting crazy (as always). Perhaps we should remove it from the Travis CI builds?

PR #86, opened 2017-04-21
Travis CI: disable pylint checks temporarily

feikname:
They're making the build status be seem as failed, even if piqueserver builds correctly.

This change should also decrease the build time.


NotAFile:
I feel like it would still be a good idea to have PyLint's input because it is great at spotting subtle bugs, but I guess the build thing just isn't the place for it.

PR #99, opened 2017-07-17
Use upstream enet

swalladge:
- removes all enet sources and associated setup scripts from piqueserver
- update dependencies to use `pyenet` from pypi
- enables installing piqueserver in develop/edit mode, since we don't have the hacks around installing enet
- should point out that the 'upstream enet' is our fork at <https://github.com/piqueserver/pyenet/> (which is the source for the pypi package as well)


NotAFile:
Just going to merge this since it LGTM and noway is MIA

noway:
LGTM 👍

noway:
The only thing I see is that some old as scripts might get broken, but that only accounts to fixing import statement in them really.

NotAFile:
Well that ship has sailed long ago really. We could still add a "from enet import *" enet.py file if we really wanted and then whack a `DeprecationWarning` in there.

noway:
It's fine not to support old import statements, as long as it is straightforward to fix.

DeprecationWarning would be nice

NotAFile:
we can however remove the sys.path hacks for enet

PR #101, opened 2017-07-17
fix /deaf

### LINKED ISSUE #100: /deaf does not work
Deaf is supposed to stop relaying chat messages from other users, but this seems to have been broken in pyspades at some point.

rcnly:
fixes #100


feikname:
That was fast.

NotAFile:
This seems to skip a bit of logic in `pyspades/server.py send_contained`. Are we sure that's not necessary?

rcnly:
oops, thanks for the pointer - it was sending teamchat to both teams! i've made another commit to fix that which i think has been added to the request automatically

pyspades/server.py
```diff
@@ -612,7 +612,10 @@ def loader_received(self, loader):
                             team = None
                         else:
                             team = self.team
-                        self.protocol.send_contained(contained, team=team)
+                        for player in self.protocol.players.values():
+                            if not player.deaf:
+                                if team is None or team is player.team:
```
swalladge: In `team is player.team`, should that be `==` rather than `is`?

swalladge: In `team is player.team`, should that be `==` rather than `is`?

NotAFile:
The major refactor I did of `pyspades/server.py` got in the way of merging this, sorry for the delay.

PR #120, opened 2017-09-04
Bump version to 0.1.0

feikname:
So we can finally release 0.1.0.

I'm unsure if we should tag the release as `0.1.0` or `0.1.0-rc1` though.


NotAFile:
I'm hesitant to release this as non-rc if I compare the radical refactorings that have been made to how little has been tested.

PR #134, opened 2017-12-03
Port to ipaddress package

NotAFile:
* Port to ipaddress package
* Add tests for networkdict

### DIFF
--- a/piqueserver/networkdict.py
+++ b/piqueserver/networkdict.py
@@ -2,9 +2,7 @@ from six import text_type
 from ipaddress import ip_network, ip_address
 
 def get_cidr(network):
-    # TODO: why are we accessing a protected attribute?
-    #       does this work?
-    if network.prefixlen == 32:  # pylint: disable=protected-access
+    if network.prefixlen == 32:
         return str(network.network_address)
     return str(network)
 
@@ -23,7 +21,7 @@ class NetworkDict(object):
         return values
 
     def remove(self, key):
-        ip = ip_network(text_type(key))
+        ip = ip_network(text_type(key), strict=False)
         networks = []
         results = []
         for item in self.networks:
@@ -36,7 +34,7 @@ class NetworkDict(object):
         return results
 
     def __setitem__(self, key, value):
-        self.networks.append((ip_network(text_type(key)), value))
+        self.networks.append((ip_network(text_type(key), strict=False), value))
 
     def __getitem__(self, key):
         return self.get_entry(key)[1]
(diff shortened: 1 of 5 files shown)


piqueserver/networkdict.py
```diff
@@ -7,19 +7,18 @@ def get_network(cidr):
     try:
         return CACHE[cidr]
     except KeyError:
-        network = IPNetwork(cidr)
+        network = ip_network(unicode(cidr))
```
NotAFile: This breaks on py3

piqueserver/networkdict.py
```diff
@@ -7,19 +7,18 @@ def get_network(cidr):
...
         CACHE[cidr] = network
         return network
 
-
 def get_cidr(network):
     # TODO: why are we accessing a protected attribute?
     #       does this work?
+    # testing for IPv4?
```
NotAFile: Enet only supports IPv4. This is testing if the network is a `/32`, i.e. is a single IP address.

piqueserver/networkdict.py
```diff
@@ -7,19 +7,18 @@ def get_network(cidr):
     try:
```
NotAFile: Do we even need get_network? Why is it caching stuff?

piqueserver/networkdict.py
```diff
@@ -54,7 +53,7 @@ def __getitem__(self, key):
         return self.get_entry(key)[1]
 
     def get_entry(self, key):
-        ip = get_network(key)
+        ip = ip_address(unicode(key))
```
NotAFile: same as above

piqueserver/networkdict.py
godwhoa: How do I make it compatible with both python 2/3?

NotAFile: It requires a unicode object, I guess? There's a few options, but it's not really easy. You could require the callee to give a valid unicode string object, alias unicode to str if unicode is not available or convert it to unicode only if it's a bytes object. We've only had this issue in reverse before.

piqueserver/networkdict.py
godwhoa: ip_network isn't an expensive call, we can remove it.

NotAFile: What I thought might be going on is that it is constanty recreating these when syncing bans

piqueserver/networkdict.py
```diff
@@ -1,20 +1,9 @@
...
-    except KeyError:
-        network = ip_network(unicode(cidr))
-        CACHE[cidr] = network
-        return network
-
 def get_cidr(network):
     # TODO: why are we accessing a protected attribute?
     #       does this work?
-    # testing for IPv4?
     if network._prefixlen == 32:  # pylint: disable=protected-access
```
NotAFile: network._prefixlen shouldn't exist in ipaddress AFAIK

NotAFile:
can you add a comment why strict=False is needed? Otherwise someone is going to be hella confused in a while.

PR #145, opened 2017-12-05
Remove commands hack

godwhoa:
Removes the hack which allowed importing via. `import commands` instead of proper `from piqueserver import commands`.
Scripts have already been rewritten to use proper import.
commands.py needed unindenting hence the ugly diff. 
Tested it locally with basic commands so no weird indentation issues.


swalladge:
Note: I've updated the wiki page on importing in scripts.

PR #156, opened 2017-12-11
Add admin_only kwarg

godwhoa:
- Added admin_only kwarg
- Re-written commands to use it
- Added basic tests for `@command`

I split up auto formatting into a different commit for the sake of clean diffs.


NotAFile:
I disagree slightly with the function signature. It should be changed to `(name=None, *aliases, admin_only=False)` so that the option is documented in the signature.

PR #163, opened 2017-12-13
Fix BytesIO import under py2

NotAFile:


swalladge:
How did this even work before!?

NotAFile:
@swalladge that's the secret it didn't

Tests just never caught it

PR #167, opened 2017-12-14
improve babel script

NotAFile:


godwhoa:
Arena script starts the countdown even if there aren't enough players. Is that fixed in this?

NotAFile:
No, this doesn't actually fix the arena script, I just noticed. It improves the code of the babel script

godwhoa:
LGTM code wise. We should test it on a server to be sure.

PR #168, opened 2017-12-14
[WIP] Refactor out gamemode

NotAFile:
This is one attempt at factoring out the gamemode. The idea is to handle the general protcol-level stuff seperately so that scripts like arena can just subclass that without needing to override a bunch of vanilla CTF behaviour.


NotAFile:
I'm just merging this since it doesn't really change anything.

PR #171, opened 2017-12-14
Increase Coverage

NotAFile:


tests/piqueserver/test_statistics.py
```diff
@@ -0,0 +1,37 @@
+"""
+test piqueserver/server.py
+"""
+
+from twisted.trial import unittest
+from unittest.mock import Mock
```
swalladge: `unittest.mock` was introduced in python3.3 - need to use something different or drop python2 support

NotAFile: that's annoying... I thought I tested it with trials

NotAFile:
Turns out I'm actually just stupid, hm. I commited the tests without commiting the fixes for bugs I wrote while testing them

NotAFile:
the fixes I wrote, not the bugs I wrote

PR #180, opened 2017-12-16
Rename color_team to team_n

### LINKED ISSUE #111: Allow using custom team names in commands.
currently, this is the way commands deal with teams:

```
    def get_team(connection, value):
        value = value.lower()
        if value == 'blue':
            return connection.protocol.blue_team
        elif value == 'green':
            return connection.protocol.green_team
        elif value == 'spectator':
            return connection.protocol.spectator_team
        raise CommandError("Invalid Team")
```

This means that these command are hard to use with custom team colors or names

NotAFile:
Closes #111

### DIFF
--- a/piqueserver/commands.py
+++ b/piqueserver/commands.py
@@ -297,14 +297,22 @@ def get_player(protocol, value, spectators=True):
 
 def get_team(connection, value):
     value = value.lower()
-    if value == 'blue':
-        return connection.protocol.blue_team
-    elif value == 'green':
-        return connection.protocol.green_team
-    elif value == 'spectator':
+    team_1 = connection.protocol.team_1
+    team_2 = connection.protocol.team_2
+    team_spectator = connection.protocol.team_spectator
+    if value == team_1.name.lower():
+        return connection.protocol.team_1
+    elif value == team_2.name.lower():
+        return connection.protocol.team_2
+    elif value == team_spectator.name.lower():
         return connection.protocol.spectator_team
-    raise CommandError("Invalid Team")
-
+    elif value == '1':
+        return connection.protocol.team_1
+    elif value == '2':
+        return connection.protocol.team_2
+    elif value == 'spec':
+        return connection.protocol.spectator_team
+    raise ValueError('Invalid Team')
 
 def join_arguments(arg, default=None):
     if not arg:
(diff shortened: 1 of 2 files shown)


piqueserver/commands.py
```diff
@@ -297,11 +297,14 @@ def get_player(protocol, value, spectators=True):
...
+    team_spectator = connection.protocol.team_spectator
+    if value == team_1.name.lower():
         return connection.protocol.blue_team
-    elif value == 'green':
+    elif value == team_2.name.lower():
         return connection.protocol.green_team
-    elif value == 'spectator':
+    elif value == team_spectator.name.lower():
         return connection.protocol.spectator_team
     raise CommandError("Invalid Team")
```
feikname: Shouldn't `ValueError()` be more appropriate?

piqueserver/commands.py
```diff
@@ -297,11 +297,14 @@ def get_player(protocol, value, spectators=True):
...
```
feikname: I believe adding additional `elif`s for `1` and `2` might be a change very convenient for those typing commands that involve teams (such as `/lock`)

pyspades/server.py
```diff
@@ -91,19 +91,19 @@ def __init__(self, *arg, **kw):
...
         self.teams = {
             -1: self.spectator_team,
-            0: self.blue_team,
-            1: self.green_team
+            0: self.team_1,
+            1: self.team_2
         }
-        self.blue_team.other = self.green_team
-        self.green_team.other = self.blue_team
+        self.team_1.other = self.team_2
```
feikname: Not that we should remove it now, but is there any particular reason this exists? Seems unnecessary

piqueserver/commands.py
NotAFile: Oh, is it? I don't know, I didn't look at this part.

pyspades/server.py
NotAFile: well, how would you programmatically access the other team otherwise? You could write an if elif thing every time, but I think this is a neat hack to shorten that.

godwhoa:
Looks good to merge once that `CommandError` issue is fixed.

NotAFile:
Has CommandError been patched?

PR #185, opened 2017-12-17
Fix random gen

NotAFile:
note @swalladge this breaks any scripts that rely on the old name of random.txt for the map script, as that shadows the global `random` module


swalladge:
Instead of renaming the map, can we simply import the map module under a unique name? 

```
diff --git a/piqueserver/map.py b/piqueserver/map.py
index 5c885d3..f9b04b3 100644
--- a/piqueserver/map.py
+++ b/piqueserver/map.py
@@ -21,6 +21,7 @@ import os
 import imp
 import math
 import random
+import uuid
 
 from pyspades.vxl import VXLData
 from piqueserver import cfg
@@ -72,7 +73,7 @@ class Map(object):
         self.load_dir = load_dir
         try:
             info = imp.load_source(
-                rot_info.name, rot_info.get_meta_filename(load_dir))
+                uuid.uuid4().hex, rot_info.get_meta_filename(load_dir))
         except IOError:
             info = None
         self.info = info
```

Here's hoping no scripts ever try to `import mapname`

NotAFile:
Yes, I agree the rename is a bad solution

NotAFile:
There are actually good changes in here too, reopening. Do not merge.

Issue #199, opened 2017-12-22
Improve command system


godwhoa:
Areas to improve:
- ~Command parsing: Currently lot of commands index `args` without checking length so if player doesn't pass in right amount of args. it results in an exception. I suggest adding `arglen` kwarg to command decorator.~

- Help/Usage messages: Right now a command has to check if proper amount of arguments are passed in and print out usage message on it's own. Standardising this would be better for both devs. and players.

NotAFile:
I disagree. The command system checks the length of the arguments of the function to see if the signature matches up.

Thus, in cases where the number of args is constant this is already implemented. The *args situation applies only to functions that can take a variable number of arguments. We can't do anything then, of course. If the function can take any number of arguments, we can't say anything about it and it has to decide for itself and raise e.g. a CommandError in case someone gets it wrong. There is just no way around that.

NotAFile:
What we could however do of course is make commands that use *args unnecessarily use the proper method of declaring the arguments in the function signature instead.

godwhoa:
> I disagree. The command system already checks the length of the arguments of the function to see if the signature matches up.

Got it. Will strikeout Command parsing part.

godwhoa:
I will closing this issue. Looks like we need to refactor the commands instead of command system itself.

NotAFile:
Might as well strike out the second part too then, since that's also implemented already :)

Look at the bottom of the commands.py file

PR #200, opened 2017-12-22
Improve core commands

feikname:
Current changes:
 * Make `/client` also work with no target player
 * Add help messages for all current core commands (#199)
 * Add a message in /script for when no scripts are enabled

### DIFF
--- a/piqueserver/core_commands/game.py
+++ b/piqueserver/core_commands/game.py
@@ -7,7 +7,7 @@ from piqueserver.commands import command, CommandError, get_player, get_team, ge
 @command('time')
 def get_time_limit(connection):
     """
-    Tells you the current time limit
+    Tell you the current time limit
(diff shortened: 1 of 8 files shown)


NotAFile:
We seem to disagree on style here a bit.

I describe what the user wants to  "Ban a player" and you describe what the command does "Bans the player". I of course think the first style reads a lot better in-game, but we should agree on the style here.

NotAFile:
Is this still WIP? Or can we target this at 0.1.1

feikname:
Feel free to merge it :)

swalladge:
Personally I prefer help messages in the imperative too. It just seems to sound better. ¯\\_(ツ)_/¯

There are also a couple of typos to be fixed once we decide on a style (descriptive vs imperative).

Issue #155, opened 2017-12-11
Redo configuration


NotAFile:
Besides containing all of the config loading logic, it would be nice if applications could declare options so they can be shown or changed at runtime (like in OpenSpades).

This could look like this:

```
enable_gravity = ConfigOption("enable_gravity", bool, default=False)
...
   enable_gravity.value
```

the `.value` does not look very nice of course, but is required as you can't change an immutable variable and the config might want to be changed at runtime.

PR #203, opened 2017-12-23
new config management

swalladge:
This builds an api for easy management and registration of configuration. It's
very much wip for now. It's following a similar design to @notafile's ideas in #155 with inspiration from [python-decouple](https://github.com/henriquebastos/python-decouple/).

Basically the idea is a global instance of the `ConfigEngine` class that
provides methods for loading and dumping configuration, as well as registering
a new configuration object with casting, defaults, and validation.


swalladge:
Basic usage is:

```py3
from piqueserver.config import config

config.load_config('some/config/file.toml')

gravity = config('gravity', cast=bool, default=True)
# gravity.value will be set to either the value of 'gravity' from the config store, or the default if not available, casted with the cast function.

print(gravity.get())

# updates the object as well as the global store
gravity.set(False)


# if a `validate` function returned false, it will raise an error
test = config('test', default='hi', validate=lambda s:len(s)<10)
```

swalladge:
Also need to decide on a syntax for accessing nested objects. Ideas:

```
config('server.passwords') # dotted
config('passwords', section='server') # section-based (but how would work for nested sections?
```

NotAFile:
RE: sections: I would make Sections their own object. Then you can do e.g. section.config("username"). Alternatively, the config can always just have a dict type and then return a dict

piqueserver/config.py
```diff
@@ -0,0 +1,86 @@
...
+        else:
+            raise Exception()
+
+    def dump_config(self, out_file=None, style='TOML'):
+        if out_file is None:
+            pprint(self.raw_config)
+        # TODO
+        pass
+
+    def get(self, name, default):
```
NotAFile: I'm kind of torn on this. On one hand a property would look a lot nicer in code, on the other hand it hides the fact that it's actually a function call.

piqueserver/config.py
```diff
@@ -0,0 +1,86 @@
...
+        # TODO: handle nested config values
+        # XXX: should requesting a config value that is a dictionary object
```
NotAFile: I believe not just returning the dictionary object will only cause arbitrary and weird restrictions in places.

piqueserver/config.py
```diff
@@ -0,0 +1,86 @@
...
+        #       return the whole dictionary, or fail?
+        return self.raw_config.get(name, default)
+
+    def set(self, name, value):
+        self.raw_config[name] = value
+
+    def __call__(self, name, cast=None, default=None, validate=None):
```
NotAFile: I think calling this `option()` or `setting()` or something would make more sense as you are declaring a config option, not a config. This would also make other types of settings, such as files or settings a lot more intuitive.

piqueserver/config.py
```diff
@@ -0,0 +1,86 @@
...
+    configuration object stored in the engine
+    '''
+    def __init__(self, engine, name, cast, default, validate):
+        self.name = name
+        self.cast = cast
+        self.default = default
+        self.validate = validate
+        self.engine = engine
+
+        value = self.engine.get(name, default)
```
NotAFile: Interesting, this is the reverse of what I would have done. I would have kept a list of ConfigObjects in the Engine and then trigger the casting and validation from there if a value is set. We have to keep in mind that for anything except scripts, the configuration is not actually available at the time of creating this object yet.

piqueserver/config.py
```diff
@@ -0,0 +1,86 @@
...
+        self.value = self._cast_and_validate(value)
+
+    def _cast_and_validate(self, value):
+        if self.cast is not None:
+            try:
+                value = self.cast(value)
+            except Exception as e:
+                print(e)
+                print('failed to cast config value')
```
NotAFile: The print is just temporary I guess

piqueserver/config.py
```diff
@@ -0,0 +1,86 @@
...
+                raise ConfigException('failed to validate')
+
+        return value
+
+
+    def get(self):
+        # TODO: use getattr magic to get and set this?
+        return self.value
+
+    def set(self, value):
```
NotAFile: I'm not sure this  is a good idea at all... I'm worried that if we add this here, people might be tempted to use config values for storage. Besides, there is no really good reason to have this. The only part of the code that should be messing with config options directly is internal anyway.

piqueserver/config.py
```diff
@@ -0,0 +1,86 @@
...
+        self.style = style
+
+        if style == 'TOML':
+            self.raw_config = toml.load(open(config_file))
+        else:
+            raise Exception()
+
+    def dump_config(self, out_file=None, style='TOML'):
+        if out_file is None:
+            pprint(self.raw_config)
```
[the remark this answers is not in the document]
swalladge: yep this was just for testing

piqueserver/config.py
swalladge: I was also thinking of something like the following to force read-only:

```
@property
def value():
  return self.value
```

(also since the value is able to be cached, the function call probably doesn't cost much)

piqueserver/config.py
swalladge: This was just for in case we wanted to make config files writeable, or to update runtime config. We need an api somewhere to update configuration.

NotAFile: well, I'm saying there's no case in which you'd actually want to edit a setting through it's `ConfigObject`. In all places where you'd have a good reason to edit the global config, you have access to the global config already.

swalladge: What about in scripts or custom commands though?

piqueserver/config.py
swalladge: aaah yeah didn't think of that....

piqueserver/config.py
NotAFile: Script writers were exactly the people I feared would abuse this...

piqueserver/config.py
```diff
@@ -0,0 +1,86 @@
...
+class ConfigEngine():
+    '''
+    configuration engine that handles loading and saving config,
+    and all the things
+    '''
+    def __init__(self):
+        self.raw_config = {}
+        self.style = 'TOML'
+
+    def load_config(self, config_file, style='TOML'):
```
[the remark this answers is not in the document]
swalladge: This could go in `run.py` where it currently is.

piqueserver/config.py
```diff
@@ -0,0 +1,145 @@
...
+import json
+import toml
+
+
+DEFAULT_STYLE = 'TOML'
+TOML_STYLE = 'TOML'
+JSON_STYLE = 'JSON'
+
+
+class ConfigException(Exception):
```
NotAFile: Do we really need a custom Exception class for this? The only point of custom exceptions is catching only a more narrow case, or giving additional information to the catchee, but I don't see how this would help here... If we need this, we should subclass ValueError though.

piqueserver/config.py
```diff
@@ -0,0 +1,145 @@
...
+class ConfigStore():
+    '''
+    configuration store that manages global configuration
+    '''
+
+    def __init__(self):
+        self._raw_config = {}
+        self.options = {}
+
+    def validate_all(self):
```
NotAFile: Does this need to be external?

piqueserver/config.py
```diff
@@ -0,0 +1,145 @@
...
+        for option in self.options.values():
+            option.validate(option.get())
+
+    def clear_config(self):
```
NotAFile: Do we ever actually want any code to call this?

piqueserver/config.py
```diff
@@ -0,0 +1,145 @@
...
+        self._raw_config = {}
+
+    def load_config(self, config_file, style=DEFAULT_STYLE):
+        self.clear_config()
+        self.update_config(config_file, style)
+
+    def update_config(self, config_file, style=DEFAULT_STYLE):
+        if style == TOML_STYLE:
+            self._raw_config.update(toml.load(open(config_file)))
```
NotAFile: I don't think we should be opening this ourselves... Instead, I think we should just pass in a file-like object. Also, this can potentially leak file descriptors as the file is never closed.

piqueserver/config.py
```diff
@@ -0,0 +1,145 @@
...
+        elif style == JSON_STYLE:
+            self._raw_config.update(json.load(open(config_file)))
+        else:
+            raise ConfigException('Unsupported config file format: {}'.format(style))
+        self.validate_all()
+
+    def load_config_object(self, config):
```
NotAFile: Object? or dict? I propose `load_from_dict`

piqueserver/config.py
```diff
@@ -0,0 +1,145 @@
...
+        self.options = {}
+
+    def validate_all(self):
+        for option in self.options.values():
+            option.validate(option.get())
+
+    def clear_config(self):
+        self._raw_config = {}
+
+    def load_config(self, config_file, style=DEFAULT_STYLE):
```
NotAFile: the config part of the name is redundant, maybe replace it with some more useful information like `load_from_file()`

piqueserver/config.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def dump_config(self, out_file, style=DEFAULT_STYLE):
+        if style == TOML_STYLE:
+            toml.dump(self._raw_config, open(out_file, 'w'))
+        elif style == JSON_STYLE:
+            json.dump(self._raw_config, open(out_file, 'w'))
+        else:
+            raise ConfigException('Unsupported config file format: {}'.format(style))
+
+    def get_raw_config(self):
```
NotAFile: I don't think we should have this. If people *really* want to access the internals, they can do it via the _raw_config attribute and will also be warned by the underscore that they are doing stupid shit and should probably overthing it.

piqueserver/config.py
```diff
@@ -0,0 +1,145 @@
...
+        return self._raw_config
+
+    def get(self, name, default=None, section=None):
```
NotAFile: This is just for backwards-compatability, right? I don't see why we would need this under the new scheme.

Also, I still want to add Section objects.

piqueserver/config.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def set(self, value):
+        if self.cast is not None:
+            value = self.cast(value)
+        self.validate(value)
+        self.store.set(self.name, value, self.section)
+
+    @property
+    def value(self):
+        return self.get()
```
NotAFile: why have this twice?

piqueserver/server.py
```diff
@@ -65,13 +66,25 @@
...
     pass
 
-# default passwords hardcoded in config
-DEFAULT_PASSWORDS = {
-    'admin': ['adminpass1', 'adminpass2', 'adminpass3'],
-    'moderator': ['modpass'],
-    'guard': ['guardpass'],
-    'trusted': ['trustedpass']
-}
+def check_passwords(passwords):
```
NotAFile: Note for future: Once we have TOML, we can just comment out the passwords section by default

piqueserver/server.py
```diff
@@ -248,7 +261,7 @@ def __init__(self, interface, config):
         self.max_players = config.get('max_players', 20)
         self.melee_damage = config.get('melee_damage', 100)
         self.max_connections_per_ip = config.get('max_connections_per_ip', 0)
-        self.passwords = config.get('passwords', {})
+        self.passwords = config_store.option('passwords', default={}, validate=check_passwords).get()
```
NotAFile: Note that this causes an error instead of just a warning

piqueserver/config.py
swalladge: Yep good idea. Btw this shouldn't leak anything though since the garbage collector will close the files on cleanup.

NotAFile: Yes, it will, in current versions of the CPython interpreter. But this is an implementation detail and can not be relied upon. In the pypy interpreter this will leak.

piqueserver/server.py
swalladge: What error does it cause?

piqueserver/config.py
swalladge: Do we want to go with `option.get()` syntax or `option.value`?

piqueserver/config.py
swalladge: I was thinking it could be useful to support backwards compatibility with scripts. What do you think?

piqueserver/config.py
swalladge: Working on section objects now. Did you have ideas for implementing it? Currently I'm thinking recursion on the main `ConfigStore` class - that would allow for infinite nesting of sections as well (not that that would be required).

swalladge: Scratch that - I don't think that will be straight forward to implement, neither will it be really useful.

swalladge: (making ConfigStore recursive that is)

piqueserver/config.py
NotAFile: I like both, but we should really decide. `.get()` might lead some "clever" script writers to "optimize" the code by only fetching it once though, thus breaking config reloading.

swalladge: That brings up another question I had: do we want to go with the flow of getting each config value direct every time it's used (so that when the config is updated, it's automatically updated everywhere it's used), or work in a command or hooks system to call reload functions when we want to reload/change the config?

One problem I can see with automatically updating config values everywhere is that in fragile parts in the server internals it might not like it if the config value changes suddenly...

NotAFile: yes, that's a concern, but those are just things we will have to iron out I'm afraid.

Luckily, the number of places the config would be changed from is pretty limited and since there is not a lot of concurrency, the parts of code that could be hit with this are limited too. The only thing that could be dangerous is with timed callbacks using `reactor.callLater()`

piqueserver/server.py
NotAFile: ẁell, in the old version it would just print a warning about default passwords. With this code, afaict, it would fail with a CommandError if any default passwords were present.

swalladge: That function returns `True`, so it will always pass the validation. I just added some clarifying comments to the code. :)

piqueserver/config.py
swalladge: I think I prefer `.get()` as it makes it more obvious that you are calling a method that is providing a value (which may not always be the same). With `.value` it feels more like a value that has no reason to change. Also we don't want anyone trying `option.value = 'something'`

swalladge:
TODO: add leading underscore to all methods and attributes that are not part of the public api.

swalladge:
TODO: investigate making `Section` recursive so it can be nested arbitarily.

swalladge:
I'm now pretty happy with the code and API for this. At the moment it is also integrated into piqueserver and available for use (I updated the protocol.passwords setting to use the new config as a proof of concept). This can perhaps be gradually integrated in until we can totally remove the previous config dictionary.

It also provides the `get_dict()` method so that we can pass it to scripts' `apply_script` function for backwards compatibility.

piqueserver/config/config.toml
```diff
@@ -0,0 +1,335 @@
...
+
+# list of messages to send when the user sends the /rules command
+rules = [
+  "Cheating isn't welcome. Griefing is frowned upon. Have fun!"
+]
+
+
+[tips]
+
+  # random lines from this list are shown at a specified frequency in-game
```
godwhoa: Is the one-space indentation required?

swalladge: no that was unintentional

swalladge: actually i think the tips should be part of the info section anyway

PR #204, opened 2017-12-23
Update spadenadefix

kmsiapps:
Fixed issues about blocking grenade when pressing both mouse buttons.


NotAFile:
I was under the impression we agreed to inline this? Either way, LGTM

kmsiapps:
I messed with commits, I'll PR again. Git's hard -_-

NotAFile:
You can just force-push instead!

PR #205, opened 2017-12-24
Fix bunch of scripts

kmsiapps:
updated spadenadefix.py
fixed `/where` (on movements.py)
fixed nointelonwalls.py

### DIFF
--- a/piqueserver/core_commands/movement.py
+++ b/piqueserver/core_commands/movement.py
@@ -77,7 +77,7 @@ def where(connection, player=None):
         raise ValueError()
     x, y, z = connection.get_location()
     return '%s is in %s (%s, %s, %s)' % (
-        connection.name, to_coordinates(x, y), int(x), int(y), int(z))
+        connection.name, to_coordinates(int(x), int(y)), int(x), int(y), int(z))
 
 
 @command('teleport', 'tp', admin_only=True)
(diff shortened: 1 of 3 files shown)


piqueserver/scripts/spadenadefix.py
```diff
@@ -1,17 +1,23 @@
...
+version 2(2017.12.23)
 """
 
+from pyspades.constants import SPADE_TOOL
 
 def apply_script(protocol, connection, config):
     class SpadenadeConnection(connection):
-
+        def on_secondary_fire_set(self, secondary):
+            self.wasSpade = (self.tool == SPADE_TOOL)
```
NotAFile: while you are at it make this `was_spade` according to pep8 perhaps

kmsiapps:
Fixes #218

piqueserver/core_commands/movement.py
```diff
@@ -77,7 +77,7 @@ def where(connection, player=None):
         raise ValueError()
     x, y, z = connection.get_location()
     return '%s is in %s (%s, %s, %s)' % (
-        connection.name, to_coordinates(x, y), int(x), int(y), int(z))
+        connection.name, to_coordinates(int(x), int(y)), int(x), int(y), int(z))
```
NotAFile: Perhaps it would be better to have to_coordinates be able to deal with floats?

kmsiapps: @NotAFile But it fixes the problem o_o Just tested it again.

Issue #215, opened 2017-12-25
IP_GETTER fails on Windows if using https


feikname:

NotAFile:
This is caused by lack of Twisted support for certificate validation on Windows.

https://twistedmatrix.com/trac/ticket/9209#ticket

Temporary workaround would be to make it use http if on windows instead

swalladge:
Can we simply make `IP_GETTER` configurable from the `config.json`? This would solve both this and the issue where `services.buildandshootcom` is down/inaccessible

swalladge:
(or even a list in the config so it can iterate through until one works)

NotAFile:
I'm going to close this since I assume this was fixed by the aiohttp switch.

PR #223, opened 2017-12-26
make ip_getter url configurable

swalladge:
In config.json:

```
"ip_getter": "http://url.tld"
```

Behaviour:

- not set: uses default ip_getter (https://services.buildandshoot.com/getip)
- set to empty string: does not query for external ip at all (useful if
you want more privacy)
- set to alternative url: uses that url to get the ip

NOTE: To work correctly, the configured url should respond with the client's
public ip address in the response body. That is, if you `curl <url>`, it
should only print out the public ip address, nothing more.

helps with #215 (though doesn't fix underlying issue)

### DIFF
--- a/piqueserver/config/config.json
+++ b/piqueserver/config/config.json
@@ -24,6 +24,7 @@
         "Cheating isn't welcome. Griefing is frowned upon. Have fun!"
     ],
     "master" : false,
+    "ip_getter" : "https://services.buildandshoot.com/getip",
     "max_players" : 32,
     "max_connections_per_ip" : 3,
     "port" : 32887,
--- a/piqueserver/config/config.json
+++ b/piqueserver/config/config.json
@@ -24,7 +24,6 @@
         "Cheating isn't welcome. Griefing is frowned upon. Have fun!"
     ],
     "master" : false,
-    "ip_getter" : "https://services.buildandshoot.com/getip",
     "max_players" : 32,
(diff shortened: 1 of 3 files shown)


swalladge:
If this is merged, we need to update the docs to include the `"ip_getter"` config option.

piqueserver/config/config.json
```diff
@@ -24,6 +24,7 @@
         "Cheating isn't welcome. Griefing is frowned upon. Have fun!"
     ],
     "master" : false,
+    "ip_getter" : "https://services.buildandshoot.com/getip",
```
NotAFile: hm, perhaps remove this from the default config? Few people will ever need to mess with it and it will just cause issues later if we change the defaults.

piqueserver/server.py
```diff
@@ -336,17 +329,25 @@ def __init__(self, interface, config):
...
 
         self.http_agent = web_client.Agent(reactor)
-        self.get_external_ip()
+
+        # ip_getter should be a url that returns solely the requester's public ip in the response body
+        # other tools:
+        # http://www.domaintools.com/research/my-ip/myip.xml
+        # http://checkip.dyndns.com/
+        # http://icanhazip.com/
+        ip_getter = config.get('ip_getter', 'https://services.buildandshoot.com/getip')
```
NotAFile: While we are at it, add an `if win32` thing here to set the default?

PR #226, opened 2017-12-26
move game modes to separate dir

### LINKED ISSUE #210: Move game modes out of script folder
Just a parallel `gamemodes` module that's the same as the `scripts` module, to reduce ambiguity. Adding this to 0.1.1 milestone as the sooner we do this the less configs we break and confusion we cause.

swalladge:
fixes #210


swalladge:
Upgrading from existing config should be the following:

1. create a `game_modes` dir in your config directory
2. move the game mode scripts you use from the scripts dir into the game_modes dir

If you were using the full module specification eg. (`piqueserver.scripts.push`), you'll need to update to (`piqueserver.game_modes.push`).

PR #232, opened 2017-12-27
inline fbpatch.py

swalladge:
This moves the logic from fbpatch.py into piqueserver/player.py. Not fully tested because I don't know how to trigger the line build exploit in question.


piqueserver/player.py
```diff
@@ -374,3 +410,14 @@ def timed_out(self):
 def encode_lines(value):
     if value is not None:
         return [encode(line) for line in value]
+
+# calculate the distance between an object with x, y, and z members and a
+# tuple containing (x, y, z) coords.
+def distance(a, b):
```
NotAFile: Very sure the Vertex3 object has this method already...

NotAFile: Yes, it does... but not directly. It implements subtraction and length. But that ammounts to the same thing.

PR #233, opened 2017-12-27
add pretty printing and error handling to ip_getter

### LINKED ISSUE #195: The error message when IP_GETTER fails is cryptic
Lets add error handling instead :)

Fallback handling would be nice too.

swalladge:
fixes #195

### DIFF
--- a/piqueserver/server.py
+++ b/piqueserver/server.py
@@ -359,10 +359,9 @@ class FeatureProtocol(ServerProtocol):
             print("Getting external IP failed:", e)
             return
 
-        self.ip = ip.exploded
-        self.ip_address = ip
-        self.identifier = make_server_identifier(self.ip, self.port)
-        print('Server public ip address: {}:{}'.format(self.ip, self.port))
+        self.ip = ip
+        self.identifier = make_server_identifier(ip, self.port)
+        print('Server public ip address: {}:{}'.format(ip, self.port))
         print('Public aos identifier: {}'.format(self.identifier))
 
     def set_time_limit(self, time_limit=None, additive=False):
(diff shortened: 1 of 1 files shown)


piqueserver/server.py
```diff
@@ -347,15 +348,23 @@ def __init__(self, interface, config):
...
         try:
             ip = yield self.getPage(ip_getter)
-        except OSError as e:
+        except Exception as e:
             print("Getting external IP failed:", e)
             return
 
-        self.ip = ip.strip()
+        ip = ip.strip()
+        if not re.match(r'^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$', ip):
```
NotAFile: I'd prefer just putting this into ip_address to outsource the validation.

Might also be worth making `protocol.ip` an IPV4Address instance... but that might cause backwards-incompatibility. Perhaps something along the lines of renaming it to `ip_address` or something and then making `ip` a property that returns `str(self.ip_address)`? The places where self.ip are used will be very limited though, so I don't think it's at all worth it.

swalladge: :+1:

I don't think protocol.ip is actually used anywhere though.

piqueserver/server.py
```diff
@@ -351,19 +350,19 @@ def get_external_ip(self, ip_getter):
...
                   'Please check the "ip_getter" setting in your config.')
             return
+        except Exception as e:
+            print("Getting external IP failed:", e)
+            return
 
-        self.ip = ip
-        self.identifier = make_server_identifier(ip, self.port)
-        print('Server public ip address: {}:{}'.format(ip, self.port))
+        self.ip = ip.exploded
```
NotAFile: The thing here is that ip and ip_address might go out of sync...

NotAFile: Also, I just looked it up and no code ever accesses `*.ip*`

PR #234, opened 2017-12-27
improve server.py file system methods

### LINKED ISSUE #91: issues with `server.py` file system methods
The fs methods at the beginning of `piqueserver/server.py` have a number of issues:

- [ ]  silently swallows errors
- [ ] does not close opened files
- [ ] very short methods that are used only once and should be inlined

swalladge:
- make sure server crashes if can't access important files
- use context manager so most open files are closed after use
- error handling for loading bans.txt

fixes #91

### DIFF
--- a/piqueserver/server.py
+++ b/piqueserver/server.py
@@ -73,6 +73,11 @@ PORT = 32887
 
 web_client._HTTP11ClientFactory.noisy = False
 
+def make_parent_dirs(filename):
+    d = os.path.dirname(filename)
+    if not os.path.exists(d):
+        os.makedirs(d)
+
 
 def random_choice_cycle(choices):
     while True:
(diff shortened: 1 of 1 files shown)


piqueserver/server.py
```diff
@@ -199,13 +182,19 @@ def __init__(self, interface, config):
...
         try:
-            self.bans.read_list(
-                json.load(open(os.path.join(cfg.config_dir, 'bans.txt'), 'r'))
-            )
-        except IOError:
-            pass
+            with open(os.path.join(cfg.config_dir, 'bans.txt'), 'r') as f:
+                self.bans.read_list(json.load(f))
+        except IOError as e:
+            if e.errno == 2: # not found
```
NotAFile: Might be more ideomatic to do
```
except FileNotFoundError:
   ...
except IOError
   ...
```

piqueserver/server.py
```diff
@@ -282,11 +271,11 @@ def __init__(self, interface, config):
         if not os.path.isabs(logfile):
             logfile = os.path.join(cfg.config_dir, logfile)
         if logfile.strip():  # catches empty filename
+            os.makedirs(os.path.dirname(logfile), exist_ok=True)
```
NotAFile: Hey, I had no idea that kwarg exists! Unfortunately, it's py3 only.

swalladge: oh that's sad... might still need that helper function then.

piqueserver/server.py
```diff
@@ -73,22 +73,10 @@
...
-def create_filename_path(path):
-    create_path(os.path.dirname(path))
-
-
-def open_create(filename, mode):
-    create_filename_path(filename)
-    return open(filename, mode)
+def make_parent_dirs(filename):
+    d = os.path.dirname(filename)
+    if not os.path.exists(d):
```
NotAFile: It's likely not an issue here, but this can cause race conditions. It's better to try create the directory and then handle the error afterwards. This could happen e.g. if two instances are started at exactly the same time.

swalladge: Hmm true true

PR #237, opened 2017-12-29
WIP: port piqueserver config to new config engine

swalladge:
- remove cfg.py
- reorganize config file structure for neatness
- default config.toml with comments
- roughly integrate with piqueserver internals to at least remove dependency on the old config.json (not fully ported yet)


swalladge:
All config ported! :tada:

Things to discuss:

- reorganizing config: some things have been moved around - need to decide whether we like it like that or need to move it back for 100% backwards compatibility with existing scripts
- backwards compatibility: can we drop support for the old config dictionary being passed to scripts?
- style of declaring and using config options (upper/lowercase, naming, resolving the value with `.get()` when assigning to the protocol attributes or resolving later, etc....)

NotAFile:
I don't personally see the point of the sections... I think all it does is make people confused

swalladge:
We'll need sections either way; it's just a matter of how we want to handle them.

piqueserver/banpublish.py
```diff
@@ -37,11 +37,11 @@ def render_GET(self, request):
 
 class PublishServer(object):
 
-    def __init__(self, protocol, config):
+    def __init__(self, protocol, port):
         self.protocol = protocol
         publish_resource = PublishResource(self)
         site = server.Site(publish_resource)
-        protocol.listenTCP(config.get('port', 32885), site)
```
NotAFile: huh, does this work?

piqueserver/config/simple.json
```diff
@@ -0,0 +1,3 @@
+{
```
NotAFile: what for? did you mean to put this in tests perhaps?

tests/piqueserver/test_config.py
```diff
@@ -137,7 +137,7 @@ def test_fail_load(self):
 
     def test_json(self):
         config = ConfigStore()
-        f = 'piqueserver/config/config.json'
+        f = 'piqueserver/config/simple.json'
```
NotAFile: yeah, I see. I don't think this should go in the config folder as that would mean it gets copied over with `--copy-config`. Ideally, just use StringIO

NotAFile:
why do we need multiple sections either way?

piqueserver/config.py
```diff
@@ -261,3 +278,8 @@ def set(self, value):
 
 
 config = ConfigStore()
+
+# commonly used config options declared here
+_temp_section = config.section('_internal')
```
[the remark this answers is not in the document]
swalladge: That was a temporary hack... I'll sort that out soon.

piqueserver/banpublish.py
swalladge: It doesn't :P oops

piqueserver/config/simple.json
swalladge: Yep it's for the tests. Agreed - using stringio would be neater for this.

swalladge:
Well for organizing I can see the following ways we could do it:

1. keep the original config layout. Backwards compatible, requires sections for nesting
2. reorganize for clarity and to keep script config in separate sections. not backwards compatible, requires sections
3. remove sections altogether and just namespace variables (eg. 'server_port' 'ban_publish_port' etc.). not backwards compatible, doesn't require sections

Personally, I think 3 is going to look pretty ugly, 1 is not much better, and 2 would be nice.

Perhaps to have less sections we could just have one big section for the core config, and some other smaller sections for ssh, irc, teams, etc.

piqueserver/banpublish.py
swalladge: Fixed

piqueserver/run.py
```diff
@@ -179,20 +188,7 @@ def main():
     if args.json_parameters:
         config.update_from_dict(json.loads(args.json_parameters))
 
-    run = True
-
-    # copy config and update geoip can happen at the same time
-    # note that sys.exit is called from either of these functions on failure
```
NotAFile: On failure? Looks like they just always get called to me.

godwhoa: hm? that's the removed comment. latest one states:
`# note these functions call sys.exit with codes 0 or 1 based on failure or success`

swalladge:
Just merged master and resolved the conflicts. Hopefully didn't break anything... :P Here's a quick todo list while I'm thinking about it:

- [ ] merge sections back together - I suggest one section for all the core config (with of course necessary sub sections or additional sections for nested config values where deemed necessary), and then we can use separate sections for each plugin/script
- [ ] remove all legacy config code
- [ ] test everything that uses the config
- [ ] anything else i forgot

We're making progress! :D

Issue #143, opened 2017-12-05
Autogenerated module Documentation


NotAFile:
Currently we have no real docs. The wiki is fine for some things I guess, but it does not allow us to generate module docs from the code or similar.

I suggest we use Sphinx to autogenerate docs, but I'm not sure there isn't something better.

If we have real docs, I'm not sure what we should do with the stuff currently in the wiki... what info should go where?

Hosting is another question. Sphinx just spits out HTML pages. Readthedocs offers free hosting for sphinx documentation, but we can also host it on GH Pages.

PR #240, opened 2017-12-30
add initial sphinx config and generated docs

swalladge:
This may or may not be what we're after, but it's a start at least.

addresses #143, but we'll have to add a lot more info to doc strings before the code is actually documented


NotAFile:
I guess we can just merge this, right? How are we going to deal with hosting the docs though?

PR #247, opened 2018-01-03
Fix irc.py and py2

godwhoa:
Changes:
Upgrade to new command system
Replace itertools.zip with six.move.zip

Replace FileExistsError with errno.EEXIST for py2 compat.
Fix kwarg in irc_say
Fix condif_dir typo in save_bans
Make filter_printable less fancy for py3 support
Fix nick encoding issue with IRC client
Fix attempted nick errors by adding nickname setter

Sorry for clumping all those changes into one commit

### DIFF
--- a/piqueserver/server.py
+++ b/piqueserver/server.py
@@ -22,6 +22,7 @@ pyspades - default/featured server
 from __future__ import print_function, unicode_literals
 import sys
 import os
+import errno
 import imp
 import importlib
 import json
@@ -95,8 +96,11 @@ def ensure_dir_exists(filename):
     d = os.path.dirname(filename)
     try:
         os.makedirs(d)
-    except FileExistsError:
-        pass
+    except OSError as e:
+        if e.errno == errno.EEXIST:
+            pass
+        else:
+            raise
 
 def random_choice_cycle(choices):
     while True:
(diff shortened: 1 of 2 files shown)


piqueserver/server.py
```diff
@@ -95,8 +96,11 @@ def ensure_dir_exists(filename):
     d = os.path.dirname(filename)
     try:
         os.makedirs(d)
-    except FileExistsError:
-        pass
+    except OSError as e:
+        if e.errno == errno.EEXIST:
+            pass
+        else:
+            raise
```
swalladge: should be `raise e`

godwhoa: @swalladge fixed the issues in latest commit can you re-review it?

swalladge: @godwhoa looks good to me. 👍

dgadelha:
@swalladge the piqueserver console might need an update then:
image

godwhoa:
@dgadelha  That one comes from [piqueserver/player.py#L350](https://github.com/piqueserver/piqueserver/blob/master/piqueserver/player.py#L349-L350). It'll get fixed in a future PR for [improving logging](https://github.com/piqueserver/piqueserver/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+logging).

NotAFile:
That's really easy to fix actually... in fact, iirc, I commited it before

Issue #251, opened 2018-01-05
ANSI escape sequences in names and messages are not stripped


NotAFile:
Seems the `str.encode("ascii", "replace")` everywhere is not useful after all.

Instead, a utility function should be created that should be used on any untrusted data before printing it to the terminal.

This function strip/replace/escape the following:
 * newlines ("\n", "\r")
 * "\x1b" (ANSI sequences)
 * Any other ASCII control sequences (anything with value under 32)

This should ideally be done with regex as to not be slow.

PR #252, opened 2018-01-06
Replace ascii encode hack with actual string escaping

NotAFile:
The goal is to go through all places where `.encode("ascii")` is used and check if using the replace function makes more sense. Really, ascii is used nowhere in piqueserver so all cases where ascii encoding is will likely have to be replaced.
  
closes:  #251


godwhoa:
Add a unit test?

NotAFile:
yeah. Are really necessary

NotAFile:
What more needs to be done?

godwhoa:
TODO: Integration testing and check if `escape_control_codes` can be used in any other place.

godwhoa:
Looks like only player.py uses acscii encode hacks. Tested it with chat messages and it works.
Good to merge.

PR #259, opened 2018-01-11
Small improvements to arena

godwhoa:
Changes: 
* Refill all players before destroying gates
* Respawn time set to 0

Tested it with two BS clients and it works.
Closes: #170


NotAFile:
What happens if spectator players exist?

godwhoa:
`arena_spawn` skips spectators so it probably causes issues. Will make `refill_all` skip spectators as well

NotAFile:
Well, I'd have just tried it... But I guess it doesn't hurt

NotAFile:
Well, I'd have just tried it... But I guess it doesn't hurt

Issue #258, opened 2018-01-11
Use csv for data collection aimbot2


godwhoa:
Currently it does something ugly:
```python
if DATA_COLLECTION:
    if self.name is not None:
        with open('aimbot2log.txt', 'a') as myfile:
            output = self.name.encode(
                'ascii', 'ignore').replace(',', '') + ','
            output += str(self.rifle_hits) + ',' + \
                str(self.rifle_count) + ','
            output += str(self.smg_hits) + ',' + \
                str(self.smg_count) + ','
            output += str(self.shotgun_hits) + ',' + \
                str(self.shotgun_count) + '\n'
            myfile.write(output)
            myfile.close()
```
Proper csv file/library should clean it up

PR #260, opened 2018-01-12
Use csv for data collection aimbot2

godwhoa:
Closes: #258


piqueserver/scripts/aimbot2.py
```diff
@@ -187,6 +184,7 @@ def hackinfo_player(player):
 
 
 def apply_script(protocol, connection, config):
+    collect_data = config.get("aimbot_collect_data", False)
```
NotAFile: have to be careful with this and the config commit

Issue #248, opened 2018-01-04
Only load config from file when running the main server


swalladge:
Currently it's loading the files before anything else (including with the --copy-config command), which makes it fail when those files don't exist yet.... (sorry!)

PR #268, opened 2018-01-15
Fix copy-config execution

godwhoa:
closes: #248

### DIFF
--- a/piqueserver/run.py
+++ b/piqueserver/run.py
@@ -49,6 +49,7 @@ def copy_config():
 
     print('Complete! Please edit the files in %s to your liking.' %
           cfg.config_dir)
+    sys.exit(0)
 
 
 def update_geoip(target_dir):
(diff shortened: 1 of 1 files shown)


godwhoa:
> Also copy-config and update-geoip should be able to be given at the same time in args.

Ah didn't think about that. LGTM.

NotAFile:
I'm really not a fan of the sys.exit(main()), and of returning 1/0. This is for a number of reasons:

1) it's not immediately obvious what those values do, their effects are way higher up in the call stack.

2) It's reinventing C-style error handling. Why? Python already has exceptions.

I agree that all sys.exit() should be as close to main() as possible, however this is taking it a step too far IMO

swalladge:
Ok that's fine. Perhaps change it to return True/False, and sys.exit(1) if `main()` returns False?

NotAFile:
I think it should raise OSError or some other exception and then catch it and raise(1), else do nothing

PR #272, opened 2018-01-16
fix /client command

### LINKED ISSUE #270: /client command is borken
babbles something about "target" not being defined

NotAFile:
* replace (broken) implicit `char *` to str conversion with decode()
 * add target back as optional arg the way it should probably originally
 have been
 * fix other stuff

closes #270


NotAFile:
This should be merged and sent out ASAP, since afaict it was in the last release.

PR #279, opened 2018-02-05
Show client info in status server

godwhoa:
Closes: #273

### DIFF
--- a/piqueserver/core_commands/player.py
+++ b/piqueserver/core_commands/player.py
@@ -13,19 +13,7 @@ def client(connection, target=None):
         player = get_player(connection.protocol, target)
         who_is = player.name + " is"
 
-    info = player.client_info
-    version = info.get("version", None)
-    if version:
-        version_string = ".".join(map(str, version))
-    else:
-        version_string = "Unknown"
-
-    return "{} connected with {} ({}) on {}".format(
-        who_is,
-        info.get("client", "Unknown"),
-        version_string,
-        info.get("os_info", "Unknown")
-    )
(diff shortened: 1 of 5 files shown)


swalladge:
Nice! Perhaps you could make `get_client_string` a method of `player`?

NotAFile:
Or, even better, a client_info data class, with a __str__ method

godwhoa:
@NotAFile That feels overkill
How about a `client_string` property of player?

swalladge:
@godwhoa perfect - the client_string only needs to be calculated once on player connect then. :+1:

godwhoa:
@swalladge Meant something like this:
```python
    @property
    def client_string(self):
        client = self.client_info.get("client", "Unknown")
        os = self.client_info.get("os_info", "Unknown")
        version = self.client_info.get("version", None)
        if version:
            version_string = ".".join(map(str, version))
        else:
            version_string = "Unknown"
        return "{} v{} on {}".format(client, version_string, os)
```

swalladge:
How about

```
def on_connect(self):
    ...
    self.client_string = self._get_client_string()

def _get_client_string(self):
   client = self..... etc.
```

godwhoa:
@swalladge client_info wouldn't be populated on connect. 
`on_version_info_recieved` would be the right hook for your idea but we also have to deal with clients that don't send version info(eg. Voxlap).
Going with `client_string` `@property` since it's more straightforward.

NotAFile:
+1, I highly prefer property over setting a string somewhere and then needing to refresh it in various places.

PR #281, opened 2018-02-11
Add tox to dev-requirements

avacore1337:
Tox was missing from dev-requirements


swalladge:
You know, technically we don't need tox in dev-requirements.txt: the deps in those files are for initing your env, but tox creates its own envs. Better off installing tox globally. :)

NotAFile:
tox uses the setup.py envs though, not the dev-requirements.txt. Thus it makes sense to have tox in there.

PR #294, opened 2018-02-18
Mergin goto into move

esplemea:
Should remove every goto to a corresponding move (if not existing already)

### DIFF
--- a/piqueserver/config/config.json
+++ b/piqueserver/config/config.json
@@ -85,8 +85,8 @@
         "trusted" : ["trustedpass"]
     },
     "rights" : {
-        "moderator" : ["advance", "cancel", "dban", "fog", "from", "goto", "hackinfo", "hban", "invisible", "ip", "kick", "kickafk", "kill", "map", "master", "move", "mute", "resetgame", "switch", "teleport", "teleport_other", "tpsilent", "togglebuild", "togglekill", "togglevotekick", "trust", "undoban", "unmute", "unstick", "where", "whowas"],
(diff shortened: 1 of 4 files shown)


godwhoa:
One more area to update:
```
piqueserver/core_commands/__init__.py:7: in <module>
    from .movement import unstick, go_to, go_to_silent, move, where, teleport, tpsilent, fly, godsilent, god, god_build
E   ImportError: cannot import name 'go_to'
```

esplemea:
oops I wanted to push on our fork, sorry. You can close this one...
My bad, we didn't finish!

NotAFile:
Hi! Thanks for the PR.

It seems you misunderstood the issue a bit. The commands /goto and /move are very similar, but not the same. So if you just rename the goto command, functionality will be missing. Also, the command is now defined twice in `movement.py` and the second definition overrides the first.

Instead, the functionality of both commands should be moved into one.

Here is what I had in mind when writing the issue:

 * `/move a4` moves you to sector a4. This is equivalent to the "goto" command
 * `/move playername a4` moves playername to a4. This is equivalent to the current "move" command.

So, way I'd do that is to look at the arguments received and decide who to move based on that. This is done in a few commands, so you can basically just copy those over.

NotAFile:
Another note: the convention is to add the s for silent to the end, not to the beginning.

Also, why close this PR?

esplemea:
Actually, we wanted to work on it on our fork and make a PR once everything's done and tested. That was a mistake I created this one so early.
Also, I choosed to put the s at the beginning since it was less confusing than "moves". But I can change that.
I will try to add also the /move <x> <y> <z> [player] but I am a bit reading for that since I am not very familiar with python and having the arguments from command lines.

PR #296, opened 2018-02-18
Replace initialize() with Team constructor calls

### LINKED ISSUE #178: Create new team instances instead of reusing the old ones
I mean, we have classes for a reason right?

Currently, teams are "reset" with team.initialize. Now, if only there was some standard way to reset a class and intialize it, Hmmmm.

albinso:
Fixes #178 

Team objects can be created based on other instances by extracting the required fields and passing them to the constructor.

Some issues with references arise when continuously creating new instances but this is covered by bundling all the reference switching into ServerProtocol.reset_teams()

### DIFF
--- a/piqueserver/game_modes/infiltration.py
+++ b/piqueserver/game_modes/infiltration.py
@@ -78,8 +78,9 @@ class DummyPlayer():
         intel_capture.winning = winning
         self.protocol.send_contained(intel_capture, save=True)
         if winning:
-            self.team.initialize()
-            self.team.other.initialize()
+            self.team = self.team(*self.team.get_init_values())
+            self.team.other = self.team(*self.team.other.get_init_values())
+            self.team.other.other = self.team
             for entity in self.protocol.entities:
                 entity.update()
             for player in list(itervalues(self.protocol.players)):
--- a/pyspades/server.py
+++ b/pyspades/server.py
@@ -230,8 +230,7 @@ class ServerProtocol(BaseProtocol):
         self.map = map_obj
         self.world.map = map_obj
         self.on_map_change(map_obj)
-        self.team_1.initialize()
-        self.team_2.initialize()
+        self.reset_teams()
         if self.game_mode == TC_MODE:
             self.reset_tc()
         self.players = MultikeyDict()
@@ -254,8 +253,7 @@ class ServerProtocol(BaseProtocol):
         player is the player which should be awarded the necessary captures to
         end the game
         """
-        self.team_1.initialize()
-        self.team_2.initialize()
+        self.reset_teams()
         if self.game_mode == CTF_MODE:
             if player is None:
                 player = list(self.players.values())[0]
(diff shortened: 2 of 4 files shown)


albinso:
We are a bit uncertain regarding the infiltration game mode. The change has been implemented and should work, but we haven't been able to configure the server to actually run the mode so it hasn't been tested in practice.

piqueserver/game_modes/infiltration.py
```diff
@@ -78,8 +78,9 @@ def score(self):
         intel_capture.winning = winning
         self.protocol.send_contained(intel_capture, save=True)
         if winning:
-            self.team.initialize()
-            self.team.other.initialize()
+            self.team = self.team(*self.team.get_init_values())
```
swalladge: Couldn't this just call `self.reset_teams()`?

swalladge: Sorry could be wrong; I'm not that familiar with the infiltration code.

NotAFile:
it's somewhat of an improvement since some attributes are new... But I agree, those values should really be fetched from the place they were initially received from.

swalladge:
Would something like this work?

```
        self.team_spectator_class = lambda: self.team_class(-1, self.spectator_name,
                                              (0, 0, 0), True, self)
        self.team_1_class = lambda: self.team_class(0, self.team1_name, self.team1_color,
                                      False, self)
        self.team_2_class = lambda: self.team_class(1, self.team2_name, self.team2_color,
                                      False, self)

        self.team_spectator = self.team_spectator_class()
        self.team_1 = self.team_1_class()
        self.team_2 = self.team_2_class()

...

def reset_teams(self):
    self.team_1 = self.team_1_class()
```
```

**Comment by NotAFile** [comment]
I don't think that's much better, tbh. I'd rather have a make_teams function that creates the teams or something. There is really no need to create teams individually at any point in the code.

**Comment by swalladge** [comment]
Rename `reset_teams` to `initialize_teams` and put all the logic in there?

File: piqueserver/game_modes/infiltration.py (hunk 1 above)
Comment (albinso) [reply piqueserver/game_modes/infiltration.py]: The infiltration code is a bit weird. This piece is from the DummyPlayer class which isn't a protocol so there is no way to use reset_teams from the ServerProtocol. Ideally resetting should be handled by the InfiltrationProtocol but that would probably require some pretty big changes in the infiltration code.

File: README.rst [hunk 2]
```diff
@@ -1,3 +1,16 @@
+Report link  
+
+https://docs.google.com/document/d/1qTW-mjOnhJ3-R_6riB_1sUDna1cKf4MjGjzGgeK9zcE/edit?usp=sharing
+
+Pull requests
```
NotAFile: Unfortunately, we can't merge this with this in there

albinso: Whoops! I forgot that the changes would show up here. I'll revert it ASAP

NotAFile:
is there still interest in making this PR mergeable?

NotAFile:
I'm going to close this for inactivity

Issue #179, opened 2017-12-16
Create new packet instances instead of reusing the old ones


NotAFile:
This might cause subtle bugs in various places in the future

Allocation costs might be an issue, of course. But luckily, cython has the `@cython.freelist` decorator to make allocation blazingly fast.

Svardl:
Hey, another KTH student group here. We're thinking about working on this issue. Could you point us to which files the issue concerns? It's not entirely clear. We found the pyspades/packet.pyx file, which seems relevant. Can we edit this file? We're unsure since it's in the pyspades directory.

NotAFile:
So basically, here's what's going on. In the top of `pyspades/server.py` and `pyspades/client.py` you will find a number of singleton instances of loaders. They are created once and then used throughout the program. This is pretty much a premature otimisation and has lead to some questionable design decisions and errors. Some packets are being filled in partially in some places and the completed in others. This is pretty intransparent. Instead, these loaders should be instantiated when the packet is actually sent and discarded afterwards. Since this means more allocation which might slow things down, the effect of this change would need to be benchmarked. If it is a noticable change, it should be looked into if this can be mitigated by using cython's freelist feature, which can be used to speed up creating new objects.

PR #297, opened 2018-02-19
Refactor: Create new packet instances instead of reusing the old ones (issue #179)

rafin:
> Create new packet instances instead of reusing the old ones  Issue #179 

All packets (except `TCState()` in `player.py`) refactored to be recreated instead of reused upon sending. Minimal testing.


pyspades/server.py
```diff
@@ -57,7 +54,7 @@ class ServerProtocol(BaseProtocol):
     connections = None
     player_ids = None
     master = False
-    max_score = 10
+    max_score = 1
```
swalladge: Was this an accidental change? :)

rafin: Yes, whoops.

NotAFile:
LGTM if you fix the typo thing.

It would be great if you made the individual commits more substantial in the future, making them this small generates a lot of noise in the logs. The solution would be to squash merge, but since there are many authors the author information would get lost and I don't really want that.

rafin:
We fixed the typo. There was also another packet initialization that we missed - should be good now.

NotAFile:
Have you tested this in-game?

rafin:
Yes, with 3 players connected simultaneously.

NotAFile:
This is starting to become too many commits for the amount of actual changes. I'm afraid we'll have to squash merge this.

rafin:
That makes sense, not an issue for us.

NotAFile:
Is this ready to be merged/tested?

rafin:
Yes it is, I'll fix the title.

NotAFile:
I guess we'll just merge and see if something breaks?

PR #298, opened 2018-02-19
Merge goto/gotos/move into move/moves

### LINKED ISSUE #112: Merge `goto` and `move` commands
These commands currently do pretty much exactly the same thing

esplemea:
Fixes #112 

The new commands are:
```
/move [player] <sector>
/move [player] <x> <y> <z>
```
And for silent move:
```
/moves [player] <sector>
/moves [player] <x> <y> <z>
```

and goto/gotos have been removed.

I made it so that if x, y or z are outside the boundaries of the world, if change them to the closest boundary.
For the 'z' coordinate: if the player was to be teleported under the ground, he is teleported on the ground instead because if he teleports inside a block, the game crashes. So this is currently impossible to teleport into a cave or under a structure because there is no easy way of telling if the position is free. Teleporting into the sky might also be dangerous as the player might die from falling afterwards.


piqueserver/core_commands/movement.py
```diff
@@ -17,49 +17,74 @@ def unstick(connection, player=None):
...
-        raise KeyError()
-    move(connection, connection.name, value, True)
+            raise ValueError()
+    do_move(connection, args)
+
+
+def do_move(connection, args, silent = False):
+    position = None
+    player = None
+    nbArgs = len(args)
```
NotAFile: This project follows the pep8 convention, so we use `snake_case`. As with the indentation thing above, you should probably run something like pylint on your files to check your files for pep8 violations and other style errors.

piqueserver/core_commands/movement.py
```diff
@@ -17,49 +17,74 @@ def unstick(connection, player=None):
...
+        z = connection.protocol.map.get_height(x, y) - 2
+        position = args[0].upper()
+    #case position is <x> <y> <z>
+    elif nbArgs == 3 or nbArgs == 4:
+        x = min(max(0, int(args[0])), 511)
+        y = min(max(0, int(args[1])), 511)
+        z = min(max(0, int(args[2])), connection.protocol.map.get_height(x, y) - 2)
+        position = '%d %d %d' % (x, y, z)
+    else:
+        return 'Wrong number of parameters!'
```
NotAFile: It would be better to raise an error here, as this automatically displays the command usage in-game.

piqueserver/core_commands/movement.py
```diff
@@ -17,49 +17,74 @@ def unstick(connection, player=None):
...
 
+    #case no player specified
+    if nbArgs == 1 or nbArgs == 3:
+        if connection not in connection.protocol.players:
+            raise ValueError()
```
NotAFile: Why is it raising an error? An error message should be added to clarify what the error encountered is.

Same goes for the other ValueError above.

esplemea: Actually I raise an error because one was already being raise before refactoring (in goto and gotos) and it's the same in the other methods in this file, such as lines 128-129. I also changed from KeyError to ValueError since it was the only place where a KeyError was thrown and it didn't make sense if it was ValueError everywhere else.

NotAFile: Oh, okay, I'll open a new issue about that then.

esplemea: Yes, also I am not sure but it's possible that sometimes no check is made at all (I mean no exception is thrown). I think this is not very consistent in this file.

piqueserver/core_commands/movement.py
esplemea: Should I raise the exception with this message instead or just raise ValueError?

piqueserver/core_commands/movement.py
NotAFile: yes, the code has a catchall in `commands.py` that displays a very unhelpful error message by default. Those should definitely be exchanged for actually helpful exceptions in the future instead.

piqueserver/core_commands/movement.py
NotAFile: Either is fine

esplemea:
So is there anything else that should be changed? I think I fixed the issues you mentioned.

PR #309, opened 2018-02-27
add method to make an untrusted set of coords safe

### LINKED ISSUE #302: Force bounds check of intel location
It can currently be dropped outside of the map.

swalladge:
- [X] add get_safe_coords() method to the map which allows quickly converting a set of coords that could be inside solid or out of bounds to a safe/sanitized set
- [ ] use on all instances where required
- [ ] make sure no out of bounds coords ever make it to `update_entities`

fixes #302


swalladge:
I'm happy with this now - it should solve #302 completely. We can work on other areas at a later date.

pyspades/server.py
```diff
@@ -338,10 +338,12 @@ def update_entities(self):
             if map_obj.get_solid(entity.x, entity.y, entity.z - 1):
                 moved = True
                 entity.z -= 1
+                # while solid in block above (ie. in the space in which the
+                # entity is sitting), move entity up)
                 while map_obj.get_solid(entity.x, entity.y, entity.z - 1):
                     entity.z -= 1
             else:
-                while not map_obj.get_solid(entity.x, entity.y, entity.z):
+                while map_obj.get_solid(entity.x, entity.y, entity.z) == False:
```
NotAFile: `is False` is preferred but that's a nitpick. Might be good to add a comment to explain that this is done since get_solid can return None

NotAFile:
maybe file the remaining stuff as issue?

PR #316, opened 2018-03-01
add more to docs

swalladge:
- move the release guide to its rightful place
- add info about testing and code cov


NotAFile:
Note one thing: in ReST, single backticks symbolize references, not code. Double backticks symbolize code.

PR #325, opened 2018-03-08
Port more things to new config engine

godwhoa:
Before I go ahead and port more scripts would like some input:

- Do we want each script to have it's own section?
- Stylistically what do you think?


swalladge:
I've of the opinion that  each script that isn't part of the core should have it's own section, even if just for namespacing.

```
[script1]
option1 = true
option2 = 3
```

vs

```
script1_option1 = true
script2_option2 = 3
```

As @NotAFile pointed out on the channel, toml supports syntax like `bcrypt = { version = "*", optional = true }` which could be useful if a script only has a few options and we want to save space in the config file.

godwhoa:
Stuff worth testing:

- [ ] rangedamage
- [ ] zoc
- [ ] spectator control
- [ ] ssh
- [ ] statusserver

godwhoa:
Testing is done!

Issue #329, opened 2018-03-16
Document script options


godwhoa:
We should have a page documenting all the options for scripts.

PR #333, opened 2018-03-30
Document script options

godwhoa:
Closes: #329


NotAFile:
Might be worth figuring out some way to autogenerate these from the scripts at some poimt

PR #337, opened 2018-06-01
port to twisted.logger

NotAFile:


NotAFile:
Note that this will mean any legacy scripts using "print" won't log anymore.

piqueserver/core_commands/server.py
```diff
@@ -1,4 +1,15 @@
+import random
+
+from twisted.logger import Logger
+
 from piqueserver.commands import command, join_arguments
+from piqueserver.config import config
+
+log = Logger()
+
+name_option = config.option(
```
swalladge: note: this is declared but not used in this file

NotAFile: Oh, this was a remnant from before I added set_server_name

NotAFile:
should be good now

godwhoa:
Neat. Can we make the log level configurable?

NotAFile:
@godwhoa yes, that would be a good next step

PR #339, opened 2018-06-01
add stricter grenade check

NotAFile:


tomlister:
Thanks.

NotAFile:
closed, discussion with @rcnly revealed that this protection was already in place, just not active by default

PR #346, opened 2018-06-07
Some cleanup/modernize work

NotAFile:


NotAFile:
merging because nobody has objected

PR #351, opened 2018-06-15
Speedup networkdict by optimizing for /32

NotAFile:
/32 and /n bans are now stored seperately, which allows for much faster
checks. Use of many range bans will still be slow though


NotAFile:
`current master: 0.033746001720428465 seconds per lookup
this PR: 0.00038178443908691405`

NotAFile:
replaced by #521

PR #355, opened 2018-07-04
add ban vacuum

NotAFile:
I don't remember how much of this worked


NotAFile:
Seems to be working fine.

PR #356, opened 2018-07-24
Improve docker image

godwhoa:
Removes source code after everything is installed.
Image size gets reduced by 71 MB.
Fixes SSL support.


NotAFile:
have you tested this, I wonder? Doesn't seem like it should work?

NotAFile:
so, how does it load the python files then? How does it load python files you haven't loaded before?

NotAFile:
ah, it runs setup.py install, I see

noway:
NICE

PR #361, opened 2018-07-27
add HELLOLAN support

### LINKED ISSUE #352: consider implementing the 1.x HELLOLAN packet
Could pretty much almost just be copied over from the ace.py code. Would anyone use it? probably not.

https://github.com/piqueserver/aosprotocol/issues/18

NotAFile:
closes #352

### DIFF
--- a/piqueserver/server.py
+++ b/piqueserver/server.py
@@ -658,15 +658,35 @@ class FeatureProtocol(ServerProtocol):
     def receive_callback(self, address: Address, data: bytes) -> None:
         """This hook recieves the raw UDP data before it is processed by enet"""
 
-        # reply to ASCII HELLO messages with HI so that clients can measure the
-        # connection latency
-        if data == b'HELLO':
-            self.host.socket.send(address, b'HI')
-            return 1
+        # exceptions get swallowed in the pyenet C stuff, so we catch anything
+        # for now. This should ideally get fixed in pyenet instead.
+        try:
+            # reply to ASCII HELLO messages with HI so that clients can measure the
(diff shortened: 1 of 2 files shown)


pyspades/server.py
```diff
@@ -316,14 +316,17 @@ def got_master_connection(self, connection):
     def master_disconnected(self, client=None):
         self.master_connection = None
 
-    def update_master(self):
-        if self.master_connection is None:
-            return
+    def get_player_count(self):
```
godwhoa: How about: 
```python
@property
def player_count(self):
```

piqueserver/server.py
```diff
@@ -658,15 +658,35 @@ def save_bans(self):
...
-        # connection latency
-        if data == b'HELLO':
-            self.host.socket.send(address, b'HI')
-            return 1
-
-        # This drop the connection of any ip in hard_bans
-        if address.host in self.hard_bans:
-            return 1
+        # exceptions get swallowed in the pyenet C stuff, so we catch anything
+        # for now. This should ideally get fixed in pyenet instead.
```
godwhoa: Make an issue for it on pyenet.

pyspades/server.py
NotAFile: I considered it. However, I don't like it when properties hide the expensiveness of calculations. In this case, I decided to make it a function since it doesn't just access the player count, but calculate it.

piqueserver/server.py
NotAFile: done

PR #364, opened 2018-07-30
add sphinx plugin to auto build config docs

swalladge:
- WIP: port rst config docs to comments in config.toml
- fix a couple of minor docs text issues
- add to sys.path in sphinx conf - hopefully this doesn't affect
readthedocs

Hoping this might help with maintaining config docs in the future (single source of truth).


swalladge:
I'll port the rest of the rst docs to the config file after any other issues are resolved and once/if generating the docs like this is approved.

NotAFile:
Hm, I don't like where this is going... At least not if config.toml is to be the default config copied into the users config directory. A number of reasons:

 - having a config file this long makes it extremely hard to navigate especially if we add long detailed descriptions to the config. As more config options get added , this will only grow worse.
- I don't think, for compactnesses sake, that every possible config option should be in config.toml. Some config options might still be valid, but very niche or rare, so we wouldn't want to bloat everyones config with them.
 - A users config will stay static to us once they do `--copy-config`, but the features and config options may change underneath. If people use the docs in the config.toml as ground truth, the comments might be invalid as soon as we release another update, with no way for people to notice changes.
 - non-essential config values should not be set unless changed, so we can change the defaults in the future without having to tell every single person with an existing config to change it.
 - the rst markup may confuse users
 - editors can't syntax highlight etc. rst in tom files. This is also true of .py files, but those contain much less documentation.

swalladge:
@notafile yeah good points. :+1:

PR #367, opened 2018-08-11
Rewrite Betterspades os info

NotAFile:
Beterspades sends the client name in the OS version as a hack because
outdated servers don't recognize it otherwise. We then double-hack and
detect that, and rewrite it back to the original text

(@xtreme8000)


NotAFile:
This should be good to merge.

Issue #345, opened 2018-06-03
Prettier config errors + storage of constructed value


NotAFile:
the config system is in and it's good, but there's a few rough edges:

- [ ] validation does not return a helpful error
- [ ] the value gets constructed every time

### Ideas, as discussed:

 - unify `validate` and `cast` into single `construct` method that returns a new value or raises an exception. This value then gets cached or stored
 - validation function returns the fields with errors and the exception object they caused. [IMO, I doesn't make to call validation implicitly anymore, at least not when assigning from a dict. It's worth discussing how manually setting an individual config option might work].

PR #368, opened 2018-08-13
Limit team names length

godwhoa:
Closes: #335


godwhoa:
`log.warn` inside validation functions don't work. Hopefully we can fix it with #345.

Issue #354, opened 2018-07-01
Inspect usage of reactor.seconds and time.time


NotAFile:
- reactor.seconds is only guaranteed to return some kind of number that increases by one every second, so it should not be used for absolute times. It currently happens to be the local unix timestamp, but this is not guaranteed.

 - time.time() returns the unix timestamp in the current timezone. This is bad because timezones change and might be different between instances. Instead, a UTC timestamp should be used everywhere.

godwhoa:
Searched for `time.time()` in codebase, all uses were for calculating time taken for an operation in seconds.
From what I've read [`time.monotonic()`](https://docs.python.org/3/library/time.html#time.monotonic) is better for that use case.

PR #370, opened 2018-08-17
Switch to time.monotonic

godwhoa:
Closes: #354


godwhoa:
Oops forgot about `reactor.seconds()`

PR #373, opened 2018-09-03
Add cmdlist command & changes to help command

godwhoa:
Changes:
 - `/help` no longer lists commands `/cmdlist` does that instead
 - `/help <command_name>` prints usage/description of a command
 - `/help` with no args prints out the help message
 -  Added a default for `help` message option, which has notes about `/cmdlist` and `/help` commands.
 -  `/cmdlist` gives a helpful output for regular players and compact output for admins

### DIFF
--- a/piqueserver/config/config.toml
+++ b/piqueserver/config/config.toml
@@ -12,16 +12,14 @@ motd = [
 ]
 
 # info displaying when user types the /help command
-# if not defined, /help will display a list of available commands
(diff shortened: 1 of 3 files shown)


godwhoa:
Did start with that initially but felt messy. Better to limit the scope of one command. UX wise I don't think it'll be an issue with the help message.

NotAFile:
How about handling it as a it a "subcommand"? E.g.

PR #375, opened 2018-10-09
Fix for map.copy()

pistoletpierre:
Fix for map not having copy() function. Alternatively, it could go right in map.py. Up to you piqueserver folks...

2018-10-09 17:27:55+0200 [-]     self.rollback_map = map.copy()
2018-10-09 17:27:55+0200 [-] AttributeError: 'builtin_function_or_method' object has no attribute 'copy'

### DIFF
--- a/piqueserver/scripts/rollback.py
+++ b/piqueserver/scripts/rollback.py
@@ -27,6 +27,7 @@ Options
 import os
 import time
 import operator
+import copy
 
 from twisted.internet.task import LoopingCall
 from pyspades.vxl import VXLData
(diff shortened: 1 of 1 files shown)


NotAFile:
Unfortunately, I don't think this will work as intended.

The `map` object is of type `VXLData`, which is ultimately implemented in C++. This means that it can't really properly copy the underlying data. Instead, it will create a new VXLData object which still points to the old map data.

NotAFile:
> Ah ok. I admittedly haven't dug at all deeply into this.

Me neither. I'm not sure what exactly it needs to copy the map for. Your fix might solve it after all.

>If it's not near the bottom on the priority list, would a deep copy or serialization+deserialization do the trick?

Deep copy wouldn't do the trick as the python runtime doesn't know what memory belongs to the VXLData object. The solution here would be to properly implement a deep copy for the VXLData object.

PR #378, opened 2018-10-10
remove client.py

NotAFile:
I think we should remove it, it's currently unused and doesn't work in the first place. If we do ever want it again, it's still in the history.


feikname:
Agreed.

PR #381, opened 2018-10-12
remove python3.4 support

NotAFile:
also bump twisted version

### DIFF
--- a/scripts/_docker_build.sh
+++ b/scripts/_docker_build.sh
@@ -9,7 +9,7 @@ set -e -x
 # code from https://github.com/pypa/python-manylinux-demo/blob/master/travis/build-wheels.sh
 # Compile wheels
 cd /io/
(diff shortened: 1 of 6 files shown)


feikname:
Why was py3.7 removed from travis for now?

NotAFile:
travis, surprisingly, doesn't support it properly yet

scripts/_docker_build.sh
```diff
@@ -9,7 +9,7 @@ set -e -x
 # code from https://github.com/pypa/python-manylinux-demo/blob/master/travis/build-wheels.sh
 # Compile wheels
 cd /io/
-for PYBIN in /opt/python/{cp27-cp27m,cp27-cp27mu,cp34-cp34m,cp35-cp35m,cp36-cp36m}/bin; do
+for PYBIN in /opt/python/{cp35-cp35m,cp36-cp36m}/bin; do
```
swalladge: do we need to add an entry similar to `cp36-cp36m` for python37?

swalladge: done in #392

PR #391, opened 2018-10-27
add develop and clean make targets

swalladge:
- `clean` to remove everything from a `make develop`
- `develop` to install inplace (same as python setup.py develop


swalladge:
There seems to be multiple ways of installing/running various things now: the Makefile, `python setup.py <command>`, `pip <command>`... We have more make targets now, but they aren't documented in the readme development instructions. Not sure what to do about it.

swalladge:
Also should be noted that the `make` commands all assume a virtualenv has already been loaded. I think this is good because we don't want to assume the location/python version of a venv, but this isn't documented yet either.

Issue #341, opened 2018-06-02
Speedhack TODO


NotAFile:
so, conclusions from talk with BR: 
 - enable speedhack detection by default
 - increase the rubberband tolerance to ~10
 - rename player.is_valid_position to e.g. player.check_speedhack or something
 - check for OOB in player.is_valid_position

NotAFile:
rubberband should also probably be a config option

PR #394, opened 2018-10-27
Set rubberband distance to 10 and make it configurable

godwhoa:
Completes TODO from issue #341 
- [x] enable speedhack detection by default (already enabled by default it seems)
- [x] increase the rubberband tolerance to ~10
- [x] rename player.is_valid_position to e.g. player.check_speedhack or something
- [x] check for OOB in player.is_valid_position (I'm guessing it is already being done in lower levels)
Closes: #341

### DIFF
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -30,6 +30,7 @@ from pyspades.mapgenerator import ProgressiveMapGenerator
 from piqueserver.config import config
 
 log = Logger()
+# distance the server tolerates between the place it thinks the client is to where the client actually is.
 rubberband_distance = config.option('rubberband_distance', default=10)
 
 
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -678,6 +678,9 @@ class ServerConnection(BaseConnection):
                 math.fabs(y - position.y) < distance and
                 math.fabs(z - position.z) < distance)
 
+    # backwards compatability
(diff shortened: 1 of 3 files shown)


pyspades/player.py
```diff
@@ -27,8 +27,11 @@
 from pyspades.common import Vertex3, get_color, make_color
 from pyspades.weapon import WEAPONS
 from pyspades.mapgenerator import ProgressiveMapGenerator
+from piqueserver.config import config
 
 log = Logger()
+rubberband_distance = config.option('rubberband_distance', default=10)
```
NotAFile: This needs documentation

pyspades/player.py
```diff
@@ -665,11 +668,11 @@ def client_string(self):
             version_string = "Unknown"
         return "{} v{} on {}".format(client, version_string, os)
 
-    def is_valid_position(self, x: float, y: float, z: float, distance: None = None) -> bool:
+    def check_speedhack(self, x: float, y: float, z: float, distance: None = None) -> bool:
```
NotAFile: This needs an alias so old scripts don't break

pyspades/player.py
NotAFile: @godwhoa it's still missing from the actual docs, i.e. the `doc/` folder

Issue #173, opened 2017-12-15
Godbuild shuld not require godmode to be activated


feikname:
It's an artificial requirement that IMO, makes no sense.

PR #396, opened 2018-10-28
Remove artificial godmode requirement for godbuild

godwhoa:
Closes: #173


godwhoa:
@feikname 
> set_god_build
Put the player into god build mode automatically when entering god mode

PR #400, opened 2018-10-30
Add unstable feature-exchange packet

NotAFile:
This would allow the client/server to check and send attributes to each other
in a standardized manner

### DIFF
--- a/piqueserver/server.py
+++ b/piqueserver/server.py
@@ -672,7 +672,7 @@ class FeatureProtocol(ServerProtocol):
             """do the actual clearing of bans"""
 
             bans_count = len(self.bans)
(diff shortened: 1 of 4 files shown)


NotAFile:
cc @xtreme8000 @yvt for comment

xtreme8000:
What would the additional data be used for? This is just to exchange feature availability, not to exchange data using an universal packet, right? I think a feature version number would suffice

NotAFile:
this is now obsolete as we have designed a new extension system.

NotAFile:
reopening so I don't forget to cherry pick the commits that accidentally landed on this branch

PR #405, opened 2018-11-01
Set up CI with Azure Pipelines

godwhoa:
Progress: Added envs. for Python 3.5, 3.6 x64. Builds are green!
Need to decide if it's worth it or not. 
Azure Pipelines lacks build caching making builds 6min(it spends 5min on installing all the dependencies). 
On the bright side it has low queue time and parallel builds.


godwhoa:
Closing for now. Might reopen once they add directory caching.

PR #411, opened 2018-11-08
Fix version request spam

godwhoa:


NotAFile:
lgtm. We should really also change the name of the send_contained method to broadcast contained.

PR #413, opened 2018-11-12
Rename send_contained to broadcast_contained.

Colton-Bouchard:


NotAFile:
Awesome! Thanks for doing this. One more thing I'd like to see before merging this is a backwards-compatible shim function that creates a DeprecationWarning, to make sure any third-party scripts don't break.

pyspades/player.py
```diff
@@ -217,7 +217,7 @@ def on_new_player_recieved(self, contained: loaders.ExistingPlayer) -> None:
                 kill_action.killer_id = player.player_id
                 kill_action.player_id = player.player_id
                 kill_action.kill_type = FALL_KILL
-                self.send_contained(kill_action)
```
godwhoa: There are two "send_contained" implementations so a full find/replace might cause some issues.
Example here: `self.send_contained` sends it to the "player" rather than broadcasting it to everyone.

godwhoa: https://github.com/piqueserver/piqueserver/blob/9884fee4c615780738d19150fd1737a532f80d57/pyspades/protocol.py#L47-L58

https://github.com/piqueserver/piqueserver/blob/96321928b755087a4142918f9faf7cf4bfab6424/pyspades/server.py#L139-L160

NotAFile:
closing due to inactivity

PR #414, opened 2018-11-15
 Add script for downloading wheels from appveyor

godwhoa:
Thought it might be useful when we need to publish windows wheels.


NotAFile:
yes, very nice

Issue #328, opened 2018-03-13
refactor auth stuff into pluggable backends


NotAFile:
The auth stuff is currently sprayed everywhere, I've locally started a seperate module to handle it.

The idea is you'd have a `AuthBackend` class you could switch. It's job is to

 * verify logins (given some input, it returns the user role)
 * return True/False for permission requests ("can this user do x")

The default one would just fetch from the config, but it's been shown by things like the aloha forum logins that there is demand for more flexible login systems.

(Besides it will remove a bunch of stuff from server.py/player.py (and a few special cases) and that's always good)

PR #415, opened 2018-11-17
Implement default config based auth backend

godwhoa:
TODO:
- [ ] redo auth limit system
- [ ] fix string formatting issues
- [ ] Figure out how it'll be pluggable(scripts?)
- [ ] Move authbackend into protocol
- [ ] Documentation for implementing a custom authbackend
- [ ] More test coverage 
- [x] "remove a bunch of stuff from server.py/player.py (and a few special cases)"

Closes: #328


piqueserver/auth.py
```diff
@@ -1,13 +1,77 @@
...
+    """Auth backend that uses the [passwords] section of the config for
     authentication"""
-    def login(self, username):
-        pass
+
+    def __init__(self):
+        self.passwords = config.option('passwords', default={}) # type: _Option
+        self.rights = config.option('rights', default={}) # type: _Option
+
+    def login(self, details: Details) -> str:
```
NotAFile: These functions must return a result immediately, any remote queries done here would hold up the whole event loop. They should be `async def` or return `Deferred` instead.

piqueserver/auth.py
```diff
@@ -1,13 +1,77 @@
...
+
+    def has_permission(self, connection, action: str) -> bool:
+        return connection.admin or action in connection.rights
+
+    def get_rights(self, user_type: str) -> List[str]:
+        rights = self.rights.get()
+        return rights.get(user_type, [])
+
+
+auth = ConfigAuthBackend()
```
NotAFile: Intuitively, I'd expect the auth backend to be an attribute of the Protocol instead. Initializing it like this makes it very hard to pass initialization parameters, or to decide which auth backend(s) should be used at runtime.

piqueserver/core_commands/social.py
```diff
@@ -10,20 +11,21 @@ def login(connection, password):
...
+        notify_login(connection, user_type)
+    except AuthError:
+        # kick if out of login attempts
+        log_failed_attempt(connection)
+        left = retries_left(connection)
+        if left < 1:
+            connection.kick('Ran out of login attempts')
+            return
+        # notify with login retries
+        return 'Invalid password - you have {} tries left'.format(left)
```
NotAFile: Similarly, I think it would be good for the AuthError to be able to specify a reason for the authentication failure to be shown to the user, in the form `raise AuthError("my reason")`

piqueserver/player.py
```diff
@@ -403,16 +403,7 @@ def on_hack_attempt(self, reason):
         self.kick(reason)
 
     def on_user_login(self, user_type, verbose=True):
-        if user_type == 'admin':
-            self.admin = True
-            self.speedhack_detect = False
```
NotAFile: Disabling speedhack detection is something that should probably be kept.

piqueserver/player.py
```diff
@@ -403,16 +403,7 @@ def on_hack_attempt(self, reason):
         self.kick(reason)
 
     def on_user_login(self, user_type, verbose=True):
-        if user_type == 'admin':
-            self.admin = True
```
NotAFile: Do any other parts of the code rely on this connection.admin attribute? It seems like they would.

NotAFile:
> Figure out how it'll be pluggable(scripts?)

My proposal here would be to have an auth backends config value, which would allow you to list the backends you want to enable. Roughly:

```toml
[auth]

backends = [
    "piqueserver.auth.ConfigAuth",
    "aloha.AlohaForumAuth",
]
```

piqueserver/player.py
godwhoa: All the logic from on_user_login has been split into `set_user_type` and `notify_login` method/function.
```python
    def set_user_type(self, connection, user_type: str) -> None:
        if user_type == 'admin':
            connection.admin = True
            connection.speedhack_detect = False
        connection.user_types.add(user_type)
        rights = set(self.get_rights(user_type))
        connection.rights.update(rights)
```

piqueserver/player.py
godwhoa: This has moved to `set_user_type` method.

NotAFile:
F

PR #434, opened 2019-01-13
Add support for new geoip database format and API version

bieito98:
Modify piqueserver/config.py and piqueserver/run.py
so download and extraction of database works
Adapt piqueserver/scripts/geoip.py to new API version

### DIFF
--- a/piqueserver/run.py
+++ b/piqueserver/run.py
@@ -4,7 +4,7 @@ import filecmp
 import shutil
 import sys
 import argparse
-import gzip
+import tarfile
 import json
 
 from piqueserver.config import (config, TOML_FORMAT, JSON_FORMAT,
@@ -58,6 +58,7 @@ def copytree(src, dst):
             else:
                 pass
 
+
 def copy_config():
     config_source = os.path.dirname(os.path.abspath(__file__)) + '/config'
     print('Attempting to copy example config to %s (origin: %s).' %
@@ -77,7 +78,7 @@ def update_geoip(target_dir):
     working_directory = os.path.join(target_dir, 'data/')
     zipped_path = os.path.join(working_directory,
                                os.path.basename(MAXMIND_DOWNLOAD))
-    extracted_path = os.path.join(working_directory, 'GeoLiteCity.dat')
+    extracted_path = os.path.join(working_directory, 'GeoLite2-City')
 
     if not os.path.exists(target_dir):
         print('Configuration directory does not exist')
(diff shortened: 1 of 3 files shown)


piqueserver/scripts/geoip.py
```diff
@@ -30,32 +40,36 @@
...
-            except KeyError:
-                continue
-            except ValueError:
-                pass
-
-            if not isinstance(value, str):
-                continue
-            items.append(value)
-        return '%s is from %s' % (player.name, ', '.join(items))
+        for raw_item in raw_items:
```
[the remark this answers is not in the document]
bieito98: Something like this?
`items = (raw_item.name for raw_item in reversed(raw_items))`
It also solves the reverse issue

piqueserver/scripts/geoip.py
```diff
@@ -1,27 +1,37 @@
...
 
 .. codeauthor:: ?
-"""
+
+Updates
+^^^^^^^
+    v1.2
+    Author: bieito98
+    Date: 13-01-2019
+    Comment: support for new GeoLite database format and API version (GeoLite2-City.mmdb & geoip2)
```
NotAFile: the formatting here isn't valid, but aside from that, I wonder it really makes sense to list the changelog for these things on the individual scripts, as opposed to the main piqueserver changelog. Thoughts?

bieito98: You are right, deleted

piqueserver/run.py
```diff
@@ -92,10 +93,12 @@ def update_geoip(target_dir):
     print('Download Complete')
     print('Unpacking...')
 
-    with gzip.open(zipped_path, 'rb') as gz:
-        d = gz.read()
-        with open(extracted_path, 'wb') as ex:
-            ex.write(d)
+    with tarfile.open(zipped_path, 'r:gz') as tar:
+        non_standard_path = os.path.join(working_directory, tar.next().name)
+        tar.extractall(working_directory)
```
NotAFile: I remembered some security issues with the tarfile program and yes: extractall on tar files is unsafe and allows attackers who control the archive (which is also currently not downloaded securely) to overwrite any file on the system. This needs to be changed.

bieito98: I see two possible solutions:

1. Check every compressed file and extract those that are safe
2. Download files via https and verify them with the md5 checksum (also downloaded securely)

In my opinion, the second one is better since it also performs an integrity check. But I don't know if that could be easily attacked.

NotAFile: I would just search the archive for the mmdb file and then extract that. The problem isn't with extracting files in general, it's with using extractall.

bieito98: Done

piqueserver/run.py
```diff
@@ -4,12 +4,14 @@
...
-import gzip
+import tarfile
 import json
 
 from piqueserver.config import (config, TOML_FORMAT, JSON_FORMAT,
-                                MAXMIND_DOWNLOAD, SUPPORTED_PYTHONS)
-import urllib.request
+                                MAXMIND_DOWNLOAD, MAXMIND_DOWNLOAD_MD5,
+                                SUPPORTED_PYTHONS)
+import requests
```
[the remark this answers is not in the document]
bieito98: Okay, updated to use `urllib.request.urlopen`

bieito98:
I didn't know it could be done like that, thanks!

Issue #412, opened 2018-11-09
Rename send_contained to broadcast_contained


godwhoa:
It's more descriptive and it'll help prevent bugs like [this](https://github.com/piqueserver/piqueserver/pull/411).

https://github.com/piqueserver/piqueserver/blob/96321928b755087a4142918f9faf7cf4bfab6424/pyspades/server.py#L139-L160

Colton-Bouchard:
I just made a PR for this issue

PR #438, opened 2019-01-19
Rename send_contained to broadcast_contained

godwhoa:
Since the other PR for #412 was stale and Colton wasn't responding to the review I decided to make this PR.
For reference: https://github.com/piqueserver/piqueserver/pull/411/commits/fabb750f508dec4cdaa4d9832b76a59b29f2599b 
Closes: https://github.com/piqueserver/piqueserver/issues/412

### DIFF
--- a/pyspades/server.py
+++ b/pyspades/server.py
@@ -136,8 +136,8 @@ class ServerProtocol(BaseProtocol):
         """alias to team_spectator for backwards-compatibility"""
         return self.team_spectator
 
-    def send_contained(self, contained, unsequenced=False, sender=None,
-                       team=None, save=False, rule=None):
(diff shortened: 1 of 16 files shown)


pyspades/server.py
```diff
@@ -159,6 +159,9 @@ def send_contained(self, contained, unsequenced=False, sender=None,
             else:
                 player.peer.send(0, packet)
 
+    # backwards compatability
+    send_contained = broadcast_contained
```
NotAFile: might be nice to have a function that generates a wrapper that creates a DeprecationWarning when called in the future.

godwhoa: That'd indeed be nice. Created an issue for it.

PR #446, opened 2019-01-21
Create issue templates

NotAFile:
I've created two issue templates since we keep asking the same things


NotAFile:
gonna merge this, we can still change them

PR #451, opened 2019-01-22
only send as many worldupdate items as we need

NotAFile:
I'm not really confident in this solution, but it appears to work. I'm particularly worried about it causing rare crashes if there is a race between the WorldUpdate and ExistingPlayer packets.

@xtreme8000 @yvt


NotAFile:
I think we determined that this was probably good to go?

xtreme8000:
I think it should be safe to use

PR #462, opened 2019-02-25
Refactor out script logic from server.py

pythonpsycho1337:


Makefile
```diff
@@ -31,3 +31,6 @@ develop:
 
 wheels:
 	./scripts/build_wheels.sh
+
+test:
```
NotAFile: note that we use `tox` for testing, which tests against a number of python versions

piqueserver/extensions.py
```diff
@@ -0,0 +1,101 @@
...
+    '''
+    seen = set()
+    dups = []
+    for script in scripts:
+        if script in seen:
+            dups.append(script)
+        else:
+            seen.add(script)
+    if dups:
+        #log.warn("Scripts included multiple times: {}".format(dups)) TODO: Pass logger object to this function
```
NotAFile: You should just create a new logger instance here. This is important because it will show the file that the message comes from. You can check existing files to see how to do this.

piqueserver/extensions.py
```diff
@@ -0,0 +1,101 @@
+
+import os
+import importlib
+import imp
+
+def check_scripts(scripts):
+    '''
+    Ensures that there are no duplicate scripts in scripts
+    :param scripts: list of scripts
+    :return: True if there is no duplicate scripts, False otherwise
```
NotAFile: we use napoleon docstring syntax: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/

piqueserver/extensions.py
```diff
@@ -0,0 +1,101 @@
...
+                script_names.remove(script)
+
+    return script_objects
+
+def apply_scripts(scripts, config, protocol_class, connection_class):
+    '''
+    Applies scripts to the specified protocol and connection class instances
+    :param scripts: List of scripts to apply
+    :param config: Config object which holds a dict
+    :param protocol_class: The protocol class  instanceto update
```
NotAFile:
```suggestion
    :param protocol_class: The protocol class  instance to update
```

tests/piqueserver/test_extensions.py
```diff
@@ -4,12 +4,22 @@
...
+from piqueserver.server import FeatureProtocol
+from piqueserver.player import FeatureConnection
 
 class TestExtensions(unittest.TestCase):
+    # Define scripts
+    script_names = ['test_script_game_mode', 'test_script_regular']
+    script_names_length = len(script_names)
+    # Get current directory
+    curr_dir = os.getcwd()
+    script_dir = os.path.join(curr_dir, 'piqueserver/test_scripts/')
```
NotAFile: sorry, I didn't see this. The test scripts should go into the `tests/` directory, not the `piqueserver/` directory, so that it does not appear in the released packages and automatic module documentation.

NotAFile: basically: `path.join(path.dirname(__file__), "test_scripts/")`

tests/piqueserver/test_extensions.py
```diff
@@ -27,24 +37,20 @@ def test_checkscripts(self):
...
     def test_load_scripts(self):
-        # Define scripts
-        script_names = ['test_script_game_mode', 'test_script_regular']
-        script_names_length = len(script_names)
-        # Get current directory
-        curr_dir = os.getcwd()
-        script_dir = os.path.join(curr_dir, 'piqueserver/test_scripts/')
+        self.assertTrue(len(self.script_objects) == self.script_names_length)
+        # apply_script method should exist in all modules
+        for i in range(0,len(self.script_objects)):
```
NotAFile: remember to follow pep8, ideally use a checker like `flake8` or a fixer like `autopep8`

Makefile
pythonpsycho1337: @arnthor89

piqueserver/extensions.py
pythonpsycho1337: Fixed!

Makefile
arnthor89: On it

piqueserver/extensions.py
NotAFile: To be more specific, we use the google style yes, since you asked in your wiki. You shouldn't add types there though, you can just leave the `(int)` off. If you really want to add types, you should [add them to the function signature instead](https://www.python.org/dev/peps/pep-0484/) to help with autocomplete. That's not very important though.

AlbertoDubbini: Ok, very helpful. We'll try to produce documentation according to that style then.
Thanks

pythonpsycho1337:
Ready for final review and merge

piqueserver/extensions.py
```diff
@@ -0,0 +1,147 @@
...
+            dups.append(script)
+        else:
+            seen.add(script)
+    if dups:
+        log.warn("Scripts included multiple times: {}".format(dups))
+        return False
+    return True
+
+
+def check_game_mode(game_mode_name):
```
NotAFile: This function seems a bit too trivial... no real logic, single layer of abstraction, only used once. While making smaller functions is generally a good idea, it's important to not overdo it, as each function you add means you need to do one more "go to definition" while reading another function. This quickly adds up in larger codebases, and you end up spending most of your time jumping around checking the implementation of things.

Here's a longer article detailing the pros and cons: https://medium.com/@copyconstruct/small-functions-considered-harmful-91035d316c29

piqueserver/extensions.py
```diff
@@ -0,0 +1,147 @@
...
+            except ImportError as e:
+                log.error(
+                    "('{} {}' not found: {!r})".format(
+                        script_type, script, e))
+                script_names.remove(script)
+
+    return script_objects
+
+
+def load_scripts_regular_extension(script_names, script_dir):
```
NotAFile: same for this wrapper

piqueserver/extensions.py
```diff
@@ -0,0 +1,147 @@
...
+        script_dir: The path to the corresponding scripts directory
+
+    Return:
+        [module]: The list of module objects containing the loaded scripts
+
+    '''
+    return load_scripts(script_names, script_dir, 'script')
+
+
+def load_script_game_mode(script_name, script_dir):
```
NotAFile: and this one

NotAFile:
whelp, should have merged this pr first... there's conflicts now

PR #464, opened 2019-02-26
Add better farbuild exploit mitigation

NotAFile:
This mitigation used to be very flakey and had many false positives.
Inspired by 1AmYF's new fbpatch2 mitigation script, which is much
better than the original fbpatch mitigation, we now record and check the
start location of the line build in addition to the end location.

This is now also more readable because these checks are actually in the correct place now, right with the other packet validity checks.

### DIFF
--- a/piqueserver/player.py
+++ b/piqueserver/player.py
@@ -125,59 +125,18 @@ class FeatureConnection(ServerConnection):
         log.info(escape_control_codes(log_message))
 
     def _can_build(self) -> bool:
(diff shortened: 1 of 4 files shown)


godwhoa:
Tested it?

NotAFile:
I haven't tested if the farbuild exploit works with this this exact patch, but line building works and the check is identical to the existing one above. I did want to try adding some unit tests though.

NotAFile:
ftr, @godwhoa tested this and it worked

PR #467, opened 2019-03-18
Add smoke test

godwhoa:


NotAFile:
I wonder if we could or should get coverage for these runs

.travis.yml
```diff
@@ -17,6 +17,7 @@ script:
   - python setup.py develop
   - pytest --cov=piqueserver --cov=pyspades
   - travis_retry coveralls
+  - ./scripts/smoke_test.py -t 25
```
NotAFile: 25s seems like a lot of time

scripts/smoke_test.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+
+
+import asyncio
+import sys
+import argparse
+
+def printer(buffer, prefix):
+    def p(data):
+        buffer.write(prefix+data)
```
NotAFile: pep8

PR #478, opened 2019-04-03
Use aiohttp for ip_getter

godwhoa:


piqueserver/server.py
```diff
@@ -169,8 +170,6 @@ def sleep(secs):
 network_interface = config.option('network_interface', default='')
 scripts_option = config.option('scripts', default=[], validate=extensions.check_scripts)
 
-web_client._HTTP11ClientFactory.noisy = False
```
NotAFile: that's actually a good question... does python's stock logging work with twisted logging ootb?

godwhoa: Nope
```
  File "test.py", line 24, in <module>
    logger.info('{ip} disconnected', ip='127.0.0.1')
  File "/usr/lib/python3.6/logging/__init__.py", line 1308, in info
    self._log(INFO, msg, args, **kwargs)
TypeError: _log() got an unexpected keyword argument 'ip'

```

NotAFile: well, that's the incorrect syntax for python's logging. My question was more if aiohttp's logs appeared in the twisted log.

godwhoa: Ah no

PR #479, opened 2019-04-03
Add new release notification

godwhoa:
`2019-04-03T20:41:09+0530 [piqueserver.version#info] New release available: v0.1.3 (Mar 1 2018): https://git.io/fjIDk`


NotAFile:
a few thoughts:
 - checking only happens on start. However, it would be a good idea to check every say 24hrs instead
 - some people hate any interaction with the outside, so it should be possible to disable via config
 - the update notification should be saved so that it can be replayed every time an admin logs in for maximum nag-factor

piqueserver/version.py
```diff
@@ -1,2 +1,30 @@
+from datetime import datetime
+from packaging import version
```
NotAFile: packaging is not a standard module

NotAFile:
Another problem I noticed here: we can't put any code that requires new dependencies into version.py, else the setup.py will fail, as version.py gets executed there to determine the current version.

PR #490, opened 2019-04-11
Basic UTF-8 encoding support

MuffinTastic:
the most simplest of changes. will probably break something terrible and obscure, but this is a good thing to have, so we should probably cross that bridge when we get there

### DIFF
--- a/pyspades/common.pyx
+++ b/pyspades/common.pyx
@@ -79,12 +79,22 @@ def crc32(data):
 # Ace of Spades uses the CP437 character set for chat and Windows-1252 for
 # player list
 
+# OpenSpades (and others) use 0xFF as a magic byte indicating the string is UTF-8 encoded
+
 def encode(value):
     if value is not None:
-        return value.encode('cp437', 'replace')
+        try:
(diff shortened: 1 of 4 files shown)


pyspades/common.pyx
```diff
@@ -79,12 +79,22 @@ def crc32(data):
...
+            return value.encode('cp437', 'strict')
+        except UnicodeError:
+            return b'\xFF' + value.encode('utf-8', 'replace')
 
 def decode(value):
     if value is not None:
+        if value[0] == 0xFF:
+            try:
+                return (value[1:]).decode('utf-8', 'strict')
+            except UnicodeError: # fallback...
```
NotAFile: This is a case only reachable with programming errors, so it should probably get a debug log line.

NotAFile:
It's worth noting that both `b'\xFF' + value.encode` and `value[1:]` copy. I don't think that we are going to have strings long enough that this becomes a performance issue though.

MuffinTastic:
yeah, thought about that.

should probably remove those unnecessary parenthesis though

PR #496, opened 2019-04-25
Detect EOF when reading VXL files

NotAFile:
Previously this just caused a very annoying segfault crash. Now we raise
a C++ exception (which cython turns into a python exceptions), which is
much more helpful.


NotAFile:
I wonder how we could unit test this without having 100MB of map files in our repo. Perhaps we could generate simple patterns on the fly... The VXL format is pretty simple.

Another thing I noticed is that it would be better and safer to just calculate the highest address instead of checking before each access. There's no point in checking every item if you can just check if there's enough space for the last one.

NotAFile:
This is really annoying because I can't reproduce this. Where is it going wrong?

Perhaps we should build with asan in CI?

PR #497, opened 2019-04-25
preliminary protocol extension support

NotAFile:
This adds support for querying the clients protocol extensions. It
doesn't do anything more, like actually send extensions, handle
extension packets, allow registering handlers for extensions or do
anything useful with the result.

### DIFF
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -5,11 +5,11 @@ import shlex
 from itertools import product
 import textwrap
 import re
+from typing import Any, Optional, Sequence, Tuple, Union, Dict
 
 from twisted.internet import reactor
 from twisted.logger import Logger
 import enet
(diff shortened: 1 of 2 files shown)


pyspades/player.py
```diff
@@ -172,6 +173,10 @@ def loader_received(self, loader: enet.Packet) -> None:
             return
         call_packet_handler(self, loader)
 
+    @register_packet_handler(loaders.ProtocolExtensionInfo)
+    def on_new_player_recieved(self, contained: loaders.ProtocolExtensionInfo) -> None:
```
godwhoa: Should be `on_ext_info_recieved` or something right?

NotAFile: oh, crap, thanks for spotting that.

Issue #500, opened 2019-04-27
Remove Piqueserver import from pyspades


NotAFile:
Commit 3a3329b9 added a `piqueserver.options` import and use into the `pyspades` package. This is bad because it makes the two packages a circular dependency and also leaks piqueservers configuration implementation into the pyspades package.

PR #501, opened 2019-04-27
Remove piqueserver.config dependency from pyspades

godwhoa:
Tried to mimic how speed_hack option was passed down.
Closes: #500


NotAFile:
Makes me think it might be a good idea to add a callback option to the options that allows you to set some function to get called every time the value gets updated.

NotAFile:
Also, nice that those things finally got `isort`ed.

PR #511, opened 2019-05-18
move logging commands to commands module

### LINKED ISSUE #473: Log irc commands
**Is your feature request related to a problem? Please describe.**
It is not possible to see the commands run from IRC in the logs

**Describe the solution you'd like**
log IRC commands, including the IRC users username

Requested/Required for Sauerkraut

swalladge:
This has the effect of supporting logging commands that are called from
irc with no code duplication.

Additionally, twisted logger is now used to log irc related information.

fixes #473

### DIFF
--- a/piqueserver/commands.py
+++ b/piqueserver/commands.py
@@ -25,12 +25,16 @@ from collections import namedtuple
 import textwrap
 from typing import Dict, List, Callable
 
+from twisted.logger import Logger
+
+from pyspades.common import escape_control_codes
 from pyspades.player import parse_command
 
 _commands = {}
 _alias_map = {}
 _rights = {}  # type: Dict[str, List[str]]
 
+log = Logger()
 
 class CommandError(Exception):
     pass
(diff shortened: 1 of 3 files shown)


piqueserver/commands.py
```diff
@@ -347,6 +351,25 @@ def get_truthy(value):
...
+    """
+    Public facing function to run a command, given the connection, a command
+    name, and a list of parameters.
+
+    Will log the command.
+    """
+    result = _handle_command(connection, command, parameters)
+
+    if result == False:
+        parameters = ['***'] * len(parameters)
```
NotAFile: This might be a good reason to introspect why this being done at all?

Also note that we need to be careful here with logging secret info such as passwords. Perhaps commands could receive a flag they can use to censor the parameters?

swalladge: Agreed. I did have a poke around and couldn't find any core commands that returned `False`. This behaviour should at least be documented.

NotAFile: Oh, so returning False is what censors the commands... I'm not convinced that's the way we should go. I think I'd personally add a decorator that sets some flag on the command function.

PR #514, opened 2019-05-19
add some player methods to irc

### LINKED ISSUE #510: Some commands crash when called through irc
**Describe the bug**

Some commands use methods of the game/server connection that aren't implemented for the irc connection, such as `send_lines`. This causes the command to crash, a traceback in the logs, and no feedback to the irc user.

**To Reproduce**

1. get a server running with irc enabled
2. call `!help` (reproducible with other commands too) from the irc channel
3. view logs

**Expected behavior**

for commands that should be available through irc: no crash

for commands that require to be called in-game: be not allowed to call it, with a feedback message.

NotAFile:
dcc6af4706 fixed these errors for ConsoleInput, but not IRC

closes #510

### DIFF
--- a/piqueserver/irc.py
+++ b/piqueserver/irc.py
@@ -188,7 +188,7 @@ class IRCBot(irc.IRCClient):
         self.send(value)
 
     def send_lines(self, lines: List[str]):
-        self.send(lines)
+        self.send("\n".join(lines))
 
 class IRCClientFactory(protocol.ClientFactory):
     protocol = IRCBot
--- a/piqueserver/irc.py
+++ b/piqueserver/irc.py
@@ -19,6 +19,7 @@ import re
 import random
 from itertools import groupby, chain
 from operator import attrgetter
+from typing import List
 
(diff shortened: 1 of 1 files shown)


NotAFile:
Note this is untested, I just did this based on swalladges's comments

swalladge:
Ultimately it would be nice to have a neater consistent api for commands to use. This passing of the connection/irc/console object to the function feels fragile.

NotAFile:
Indeed... but I'm not quite sure how else you'd do it? Perhaps some kind of `CommandCaller` wrapper object that you'd have to call `get_connection()` on to get a player reference? That would then err out and say "This command must be called from a player" or something?

I'm also not sure how you could change this in a backward-compatible manner.

NotAFile:
Actually, scratch that... I wanted to make a new command API anyway! Wanted to get rid of global command registration in favor of multiple `CommandRegistry` objects. This would allow you to group commands for better documentation and rights management, as well as bringing us closer to being able to load and unload scripts at runtime.

PR #515, opened 2019-05-19
Split permissions for moving and moving others

### LINKED ISSUE #474: Merging of goto and move removes scenario where players are allowed to move themselves but not others
previously, by granting `goto` but not `move`, it was possible to allow players to move only themselves on e.g. build servers. This is no longer possible.

NotAFile:
the move_others or admin right is now required to move other players

closes #474

### DIFF
--- a/piqueserver/core_commands/movement.py
+++ b/piqueserver/core_commands/movement.py
@@ -1,5 +1,6 @@
 from pyspades.common import (coordinates, to_coordinates)
-from piqueserver.commands import command, CommandError, get_player
+from piqueserver.commands import (command, CommandError, get_player,
+                                  PermissionDenied)
 
 
 @command(admin_only=True)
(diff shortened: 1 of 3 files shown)


piqueserver/core_commands/movement.py
```diff
@@ -74,6 +79,9 @@ def do_move(connection, args, silent=False):
         player = connection.name
     # player specified
     elif arg_count == 2 or arg_count == 4:
+        if not connection.rights.admin or connection.rights.move_others:
```
godwhoa: Shouldn't this be `if not (connection.rights.admin or connection.rights.move_others):`

NotAFile: good catch yes

NotAFile:
```suggestion
        if not (connection.rights.admin or connection.rights.move_others):
```

PR #516, opened 2019-05-19
add the user to automatic rights groups

### LINKED ISSUE #475: pseudo-roles for all, team_1, team_2
**Is your feature request related to a problem? Please describe.**
It is not possible to grant permissions to everyone or individual teams without modifying the server code.

**Describe the solution you'd like**
`all`, `team_1` and `team_2` pseudo-roles in the config that assign the relevant permissions to those people.

**Additional context**
Sauerkraut would like this.

NotAFile:
add the team's user type and also a generic all user type

This allows you to give permissions to run certain, usually restricted
commands to anyone or to a certain team

closes #475

### DIFF
--- a/piqueserver/player.py
+++ b/piqueserver/player.py
@@ -72,6 +72,9 @@ class FeatureConnection(ServerConnection):
                 self.disconnect(ERROR_BANNED)
                 return
 
+        self.user_types = pyspades.types.AttributeSet()
+        self.rights = pyspades.types.AttributeSet()
+
         ServerConnection.on_connect(self)
 
     def on_join(self) -> None:
@@ -88,8 +91,6 @@ class FeatureConnection(ServerConnection):
         self.protocol.irc_say('* %s (IP %s, ID %s) entered the game!' %
                               (self.name, self.address[0], self.player_id))
         if self.user_types is None:
-            self.user_types = pyspades.types.AttributeSet()
-            self.rights = pyspades.types.AttributeSet()
             if self.protocol.everyone_is_admin:
                 self.on_user_login('admin', False)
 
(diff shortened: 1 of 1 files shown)


piqueserver/player.py
```diff
@@ -265,8 +266,38 @@ def on_team_join(self, team: 'FeatureTeam') -> HookValue:
...
                 self.send_chat('Team is full, moved to %s' % other_team.name)
                 return other_team
+
+        # Add the user to their relevant groups
+        self.user_types.add("all")
+
+        team_usertype = None
+
+        if team is self.protocol.team_1:
+            team_usertype = "team_1"
```
swalladge: Perhaps we could use `team1` and `team2` to be consistent with the toml config sections?

piqueserver/player.py
```diff
@@ -265,8 +266,38 @@ def on_team_join(self, team: 'FeatureTeam') -> HookValue:
...
+        elif team is self.protocol.team_2:
+            team_usertype = "team_2"
+        elif team.spectator:
+            team_usertype = "spectator"
+
+        if team_usertype:
```
swalladge: If no `team_usertype`, shouldn't all the special usertypes still be removed?

piqueserver/player.py
```diff
@@ -88,8 +91,6 @@ def on_login(self, name: str) -> None:
         self.protocol.irc_say('* %s (IP %s, ID %s) entered the game!' %
                               (self.name, self.address[0], self.player_id))
         if self.user_types is None:
```
swalladge: Change to `if not self.user_types:` since it might be an empty set rather than None.

swalladge:
Something else to consider: server config shouldn't allow giving passwords for built-in roles such as `team_1`, and users should not be able to `/login team_1`

piqueserver/player.py
NotAFile: That's a good point. Kind of unfortunate that our config and code differ now.

piqueserver/player.py
NotAFile: That's true. team_usertype will never be None in the forseeable future, so the `if` is pretty pointless anyway.

NotAFile:
> server config shouldn't allow giving passwords for built-in roles such as `team_1`, and users should not be able to `/login team_1`

Why not? I don't see any harm in this technically being possible.

piqueserver/player.py
NotAFile: Hm, you're right, we have a problem here. But it's a different one. `on_join` is called multiple times for the lifetime of a connection, once per match. We probably don't want to login the users as admin every time when a new match starts. But we need the user to be joined to be able to execute `on_user_login` as it might send chat messages.

swalladge:
> Why not? I don't see any harm in this technically being possible. It should work fine, but it'd be pretty pointless... I don't think we need to protect our users from being able to do something pointless?

I guess. I was thinking just a check that rejected logging in with a message about it being an automatic/dynamic role.

PR #521, opened 2019-05-21
Make Ban Checking Constant Time

### LINKED ISSUE #158: Make ban checking fast again
Currently, ban checking is very slow as every IP is checked against every single ban entry. On servers with around 20k bans to check like the aloha servers, this is a very real performance issue. The Server takes dozens of milliseconds to evaluate each joined player, potentially causing lag. It would be better if bans were sorted, allowing almost instant checks.

There are external C modules that make this very fast like `sortedcollections`, but it was also possible to achieve good performance in my tests with the built-in bisect module. In this case, insertions are much slower though, since python needs to move a worst case of n elements if an IP is inserted in the front. In my testing, that time was very low, but with ban synchronization several thousand bans might be inserted at once so it might become an issue.

NotAFile:
Ban checking is now constant time. This makes ban checking for a small
banlist slower by a factor of 10. However, more importantly, ban
checking now takes the same amount of time regardless of how many bans
you have stored.

 Original Pyspades

10 8.670362993143498e-06
100 5.076720300712623e-05
1000 0.0004664193200005684
10000 0.004667077518010047

 New Constant Time

10 9.937950099993032e-05
100 0.00010137572599342092
1000 0.00010073935400578194
10000 0.00010321205800573807

It would be very easy to make this faster even for lower values.
IPv4Address objects are heavy to create, and we create a bunch of them.
They could be created only when necessary for output, and a faster (ip:
int, mask: int) tuple could be used internally.

However, this commit solves the problem aloha is having and I have to go
now.

closes #158

### DIFF
--- a/piqueserver/networkdict.py
+++ b/piqueserver/networkdict.py
@@ -1,4 +1,5 @@
 from ipaddress import ip_network, ip_address
+from collections import OrderedDict
 
 def get_cidr(network):
     if network.prefixlen == 32:
@@ -10,7 +11,7 @@ def get_cidr(network):
 
 class NetworkDict(object):
     def __init__(self):
-        self.networks = []
+        self.networks = OrderedDict()
 
     def read_list(self, values):
         for item in values:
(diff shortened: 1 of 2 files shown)


NotAFile:
Ah, one node, we lose the ability to have multiple ban entries for one IP. I assume this is acceptable.

Issue #105, opened 2017-08-12
Get rid of `pyspades/types.py`


NotAFile:
This file consists almost exclusively of types that are completely pointless.

One example is this gem:
```python
class AttributeSet(set):
    """
    set with attribute access, i.e.

        foo = AttributeSet()
        foo.bar = True
        foo.bar == 'bar' in foo == True
        foo.bar = False
        foo.bar == 'bar' in foo == False

    This works as a quick shorthand for membership testing.
    """

    def __getattr__(self, name):
        return name in self

    def __setattr__(self, name, value):
        if value:
            self.add(name)
        else:
            self.discard(name)
```

NotAFile:
The worst offender here is `MultiKeyDict`, which essentially turns the nice probable O(1) of a dictionary into an O(n) search every time you want to get a player by name or ID. The authors rationale was aparently not needing two dicts, but it's fair to say that this is not a good design decision.

swalladge:
All useless types removed. :)

PR #522, opened 2019-05-27
remove MultikeyDict

swalladge:
ref #105

This seems to be removable without any ill affects, as the multikey
functionality doesn't appear to be used anywhere in the repo.

Not sure about the other two classes defined in pyspades/types.py -
IDPool seems useful and AttributeSet is widely used (and this change
wouldn't be backwards compatible for third-party scripts that use it).


NotAFile:
Agreed regarding the other types.

PR #523, opened 2019-05-27
add tests to import previously unimported modules

swalladge:
simple smoke test to catch syntax errors, import errors, etc.

Also caught a couple of issues:

- missing import
- `conch` feature required for twisted to use ssh...

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -62,6 +62,17 @@ pip (stable version)
 
     pip3 install piqueserver
 
+Optional features:
+
+- `ssh`: enable ssh manhole server support
+- `from`: enable the `from` command to geolocate players by ip
+
+To install with optional features with pip:
+
+.. code:: bash
+
+    pip3 install piqueserver[ssh,from]
+
 git (bleeding edge)
 ~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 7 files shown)


NotAFile:
I still have tests for ssh lying around... they require external files though and I haven't fixed that yet.

re conch, we have a piqueserver[ssh] extra dep, we should probably check the dependencies of that and either remove the extra, or add twisted[tls,conch] there.

swalladge:
>  we have a piqueserver[ssh] extra dep

We _did_ until recently. I removed it in 49bffaa because the deps we had for `[ssh]` were already part of twisted[tls]. Turns out that conch.ssh also requires extras like `bcrypt` that weren't previously apparent.

I guess our options are:

1. include `twisted[tls,conch]` by default
2. reinstate the import checking in ssh.py and readd a `[ssh]` extra deps section (but obviously depending on the correct deps this time)

NotAFile:
hm, not sure what's the best option here

swalladge:
Something to consider is that it adds `bcrypt` and `appdirs` to the dependencies, and we are already talking about including appdirs by default. I was going to say just include it by default to make things simpler, but we do still have the optional deps for the `from` command, so optional deps still exist. Either way.

NotAFile:
We might also kind of want `bcrypt` for password hashing if/when we add that.

swalladge:
How about for now we add back the optional [ssh]. At least that's what users have been used to for a long time.

PR #525, opened 2019-05-28
document and fix everyone_is_admin functionality

### LINKED ISSUE #295: Config option to make all players admin
useful for testing stuff

swalladge:
Previously this would only be activated upon `/reloadconfig`. Now it's moved into the core so it will be activated from the start. 

fixes #295

### DIFF
--- a/piqueserver/config/config.toml
+++ b/piqueserver/config/config.toml
@@ -171,7 +171,7 @@ cap_limit = 10
 
 # Set this to true to automatically log all players in as admin on join.
 # Possibly useful for testing purposes.
-everyone_is_admin = true
+#everyone_is_admin = false
 
 [team1]
 # information about the team1


piqueserver/config/config.toml
```diff
@@ -196,6 +196,10 @@ color = [ 0, 255, 0]
 #guard = ["guardpass"]
 #trusted = ["trustedpass"]
 
+# Include an empty password for the admin role to automatically log in everyone
+# as admin. Could be useful for testing.
+#admin = [""]
```
NotAFile: I don't like this, as it's implicit and could happen accidentally, which is why we removed this behavior initially. I'd prefer a designated everyone_is_admin option. I thought we already had that, but seems not.

swalladge: i'm not sure we ever did remove the behaviour. it was still there, just one /reloadconfig away...

PR #526, opened 2019-05-29
Update bansubscribe format and fix operation

### LINKED ISSUE #483: Ban Subscribe appears to be broken
**Describe the bug**
kuunikal says bansubscribe did not work for him

### LINKED ISSUE #519: Change Bansubscribe Config Format
The current Bansubscribe format is:
```
[[url, [filternames, ...], ...]
```

so

```
List[List[str, List[str]]]
```
which is hard to understand.

We should have another format, that's more intuitive and better representable in TOML instead

Something like:

```
[[bans.bansubscribe]]
url = "https://..."
filter = ["name", "name2"]

[[bans.bansubscribe]]
url = "https://"
```

Or the equivalent:

```
[bans]
bansubscribe = [
    { url = "https://...", filter = ["name", "name2"]},
    { url = "https://..."},
]
```

### LINKED ISSUE #524: bansubscribe urls format invalid in toml
Toml only allows homogeneous lists, while the type is currently `[[str, list], [str, list]]`. It also doesn't seem to like tuples (otherwise `[(str, list)]` could be valid).

**To Reproduce**

Uncomment the example urls value in the config. See https://github.com/piqueserver/piqueserver/blob/894d8521c04ebd1db4be23043da0bc99df28f72e/piqueserver/config/config.toml#L217-L219

swalladge:
note: changes to config.py are to make sure check_unused works
correctly. probably will need some more tweaks in future too.

This is a backwards incompatible change for the bansubscribe config
format, but it was previously broken anyway, so no-one has probably been
using it.

fixes #524
fixes #519
fixes #483

### DIFF
--- a/doc/config-options.rst
+++ b/doc/config-options.rst
@@ -137,7 +137,11 @@ This section defines the behaviour when admins ban players::
     publish_port = 32885
 
     # Bansubscribe allows you to inherit bans from another server with banpublish enabled.
-    urls = [ [ "http://www.blacklist.spadille.net/subscribe.json", []]]
+    # `url` is the URL returning the json list, `whitelist` is a list of names which should
+    # be exempt from the filter
+    bansubscribe = [
+        { url = "http://www.blacklist.spadille.net/subscribe.json", whitelist = []},
+    ]
 
 .. _respawn_waves:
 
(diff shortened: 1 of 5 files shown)


swalladge:
@Kuunikal it would be great if you could try this branch to see if it fixes the bansubscribe for you. :)

PR #532, opened 2019-08-09
Little less trust in the client side

Jipok:


pyspades/world_c.cpp
```diff
@@ -506,6 +506,11 @@ inline void set_orientation_vectors(Vector *o, Vector *s, Vector *h)
 
 void reorient_player(PlayerType *p, Vector *orientation)
 {
+    Vector *o = orientation;
```
NotAFile: I'd add a `.normalize()` and `.length()` method to `Vector` instead. Variable names are also not good. Also, I really don't like this code. `o` never gets used, but it is aliased to `orientation` and modified through that instead. This kind of trickery should be avoided to keep code understandable.

pyspades/player.py
```diff
@@ -336,6 +338,17 @@ def on_input_data_recieved(self, contained: loaders.InputData) -> None:
...
+        # Do not let jump if the player is in the air
+        check = self.protocol.map.get_solid
+        pos = world_object.position
+        oz = 2 + int(not contained.crouch) # TODO CHECCK THIS        
+        if contained.jump and not (
+        check(pos.x+0.5, pos.y+0.5, pos.z + oz) or
+        check(pos.x+0.5, pos.y-0.5, pos.z + oz) or 
+        check(pos.x-0.5, pos.y+0.5, pos.z + oz) or 
+        check(pos.x-0.5, pos.y-0.5, pos.z + oz)):
+            print("WARNING: Attempt to take off. " + str(self.name))
```
NotAFile: Don't use print, do use the built in logging. Also, use string formatting instead of +

pyspades/player.py
```diff
@@ -336,6 +338,17 @@ def on_input_data_recieved(self, contained: loaders.InputData) -> None:
         z_vel = world_object.velocity.z
         if contained.jump and not (z_vel >= 0 and z_vel < 0.017):
             contained.jump = False
+        # Do not let jump if the player is in the air
+        check = self.protocol.map.get_solid
+        pos = world_object.position
+        oz = 2 + int(not contained.crouch) # TODO CHECCK THIS        
+        if contained.jump and not (
```
NotAFile: This is not valid pep8 formatting

pyspades/world_c.cpp
```diff
@@ -506,6 +506,11 @@ inline void set_orientation_vectors(Vector *o, Vector *s, Vector *h)
...
+    float d = sqrtf(o->x * o->x + o->y * o->y + o->z * o->z);
```
NotAFile: don't use the old sqrtf C function, use `std::sqrt`

pyspades/player.py
```diff
@@ -222,6 +222,8 @@ def on_orientation_update_recieved(self, contained: loaders.OrientationData) ->
             self.on_hack_attempt(
                 'Invalid orientation data received')
             return
+        # WARNING: Non-standard client can send a non-normalized vector
```
NotAFile: why not do it earlier then?

Jipok:
Here we need a full refactoring of the code related to physics. I have no desire to do this now. I leave this pr open

Issue #299, opened 2018-02-20
Make sure all error conditions in commands have messages


NotAFile:

NotAFile:
Basically what this involves is taking all of the
```
raise ValueError()
```

stuff in the commands and replacing it with

```
raise ValueError("text describing the error")
```

PR #543, opened 2019-08-14
Refactor commands

godwhoa:
Goals thus far:
* Eliminate ValueError()
* DRY up commands like /fly [player]

closes:  #299

### DIFF
--- a/piqueserver/commands.py
+++ b/piqueserver/commands.py
@@ -427,8 +427,8 @@ def _handle_command(connection, command, parameters):
         msg = str(e)
     except PermissionDenied as e:
         msg = 'You can\'t do that: {}'.format(str(e))
-    except ValueError:
-        msg = 'Invalid parameters'
+    except ValueError as e:
+        msg = str(e) if e.args else "Invalid parameters"
 
     return format_command_error(command_func, msg)
 
--- a/piqueserver/commands.py
+++ b/piqueserver/commands.py
@@ -268,10 +268,33 @@ def player_only(func: Callable):
         func(connection, *args, **kwargs)
     return _decorated
 
+def target_player(func: Callable):
+    """This decorator converts first argument of a command to a `piqueserver.FeatureConnection`.
+       It's intended for commands which accept single argument for target player eg. /fly [player].
+       It implicitly uses invoker as target if no arguments are provided.
+       It uses first argument are player name or id for targetting.
+       It forces non-player invokers to provide player argument.
 
-# TODO: all of these utility functions should be seperated from the actual
-# implementation of the commands
-
+    >>> @command()
+    ... @target_player
+    ... def fly(connection, target):
+    ...     target.fly = True
+    ...     pass
+    """
+    @functools.wraps(func)
+    def _decorated(connection, *args, **kwargs):
+        is_player = connection in connection.protocol.players.values()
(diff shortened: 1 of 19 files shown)


godwhoa:
All ValueError()s removed. target_player got rid of a lot of them, but I haven't manually tested the affected commands.
If we can test out couple of the affected commands this PR should be good to merge.

piqueserver/core_commands/info.py
```diff
@@ -13,21 +14,16 @@ def streak(connection):
...
-        if connection not in connection.protocol.players.values():
-            raise ValueError()
-        player = connection
-    else:
-        player = get_player(connection.protocol, value)
     ping = player.latency
-    if value is None:
-        return 'Your ping is %s ms. Lower ping is better!' % ping
-    return "%s's ping is %s ms" % (player.name, ping)
+    if connection == player:
```
NotAFile: you really want to check using `is` here, as you are checking identity, not equality

piqueserver/scripts/runningman.py
```diff
@@ -61,18 +61,12 @@ def relink(connection):
 
 
 @command(admin_only=True)
+@target_player
 def unlink(connection, player=None):
```
swalladge: Player doesn't need to default to None here anymore either.

piqueserver/commands.py
```diff
@@ -268,10 +268,33 @@ def _decorated(connection, *args, **kwargs):
...
+    """
+    @functools.wraps(func)
+    def _decorated(connection, *args, **kwargs):
+        is_player = connection in connection.protocol.players.values()
+        # implicitly set target to invoker if no args
+        if len(args) == 0 and is_player:
+            args = [connection]
+        # try and use first arg as player name or id to target
+        elif len(args) > 0:
+            args = [get_player(connection.protocol, args[0]), *args[1:]]
```
swalladge: Possibly clearer to do this?:

```
args[0] = get_player(connection.protocol, args[0])
```

godwhoa: It would be, but args is a tuple and you can't reassign items in it.

piqueserver/scripts/runningman.py
godwhoa: Ah true, forgot to remove in a couple of places.

swalladge:
oh right lol, fair enough then. Maybe arcs should be kept as a tuple then for consistency?

August 17, 2019 8:31:50 AM UTC Joseph Daniel <notifications@github.com> kirjoitti:
>godwhoa commented on this pull request.
>
>
>
>> -
>+    >>> @command()
>+    ... @target_player
>+    ... def fly(connection, target):
>+    ...     target.fly = True
>+    ...     pass
>+    """
>+    @functools.wraps(func)
>+    def _decorated(connection, *args, **kwargs):
>+        is_player = connection in connection.protocol.players.values()
>+        # implicitly set target to invoker if no args
>+        if len(args) == 0 and is_player:
>+            args = [connection]
>+        # try and use first arg as player name or id to target
>+        elif len(args) > 0:
>+            args = [get_player(connection.protocol, args[0]),
>*args[1:]]
>
>It would be, but args is a tuple and you can't reassign items in it.

-- 
Envoyé de mon appareil Android avec Courriel K-9 Mail. Veuillez excuser ma brièveté.

PR #544, opened 2019-08-17
Replace imp with importlib

godwhoa:
Separate script finding and loading for better errors

### DIFF
--- a/piqueserver/extensions.py
+++ b/piqueserver/extensions.py
@@ -1,7 +1,6 @@
 
 import os
 import importlib
-import imp
 
 from twisted.logger import Logger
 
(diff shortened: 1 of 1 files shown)


piqueserver/extensions.py
```diff
@@ -63,25 +62,24 @@ def load_scripts(script_names, script_dir, script_type):
...
+        if not spec:
+            log.error(
+                "{} '{}' not found in either {} directory or global scope".format(
+                    script_type, script, script_dir))
+            continue
         try:
-            # this finds and loads scripts directly from the script dir
-            # no need for messing with sys.path
-            f, filename, desc = imp.find_module(script, [script_dir])
-            module = imp.load_module('piqueserver_{}_namespace_'.format(
```
NotAFile: the removal of the namespace modification will break script names that shadow global module names again :/

PR #555, opened 2019-12-01
0.76 support

iamgreaser:
Currently a WIP.

This adds support for the 0.76 protocol. There's also some cleanups of the weapon stuff as we're doing a lot of stuff to pyspades/weapons.py here anyway.

A goal is to eventually be able to have clients speaking different protocols at the same time, which will be useful for a potential overhaul of the protocol as it stands. I don't promise that it will be a good idea to run any pair of protocol or game versions at the same time, but 0.75 and 0.76RC10 should be close enough to each other gameplaywise to not be absolutely awful.

### DIFF
--- a/pyspades/packet.pyx
+++ b/pyspades/packet.pyx
@@ -17,11 +17,14 @@
 
 from pyspades.loaders cimport Loader
 from pyspades.bytes cimport ByteReader, ByteWriter
+from pyspades.constants import GAME_VERSION_COUNT
 
 _client_loaders = {}
 _server_loaders = {}
 
-def register_packet(loader=None, server=True, client=True, extension=None):
+def register_packet(loader=None,
+                    server=True, client=True,
+                    extension=None):
     """register a packet
 
     >>> @register_packet()
@@ -66,8 +69,8 @@ def register_packet(loader=None, server=True, client=True, extension=None):
     # lifetime, but only time will tell
 
     def register(cls):
-        # Having to handle the two loader dicts separate isn't very nice, but I
-        # you can't avoid it whithout making things more complicated somewhere
+        # Having to handle the two loader dicts separate isn't very nice, but
+        # you can't avoid it without making things more complicated somewhere
         # else.
         # This is needed because certain packets (e.g. HitPacket and SetHP) in
         # likely a legendary case of premature optimisation (or because the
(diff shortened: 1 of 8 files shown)


NotAFile:
The tests appear to fail because the assert you added is hit by `test_team_join`.

iamgreaser:
> The tests appear to fail because the assert you added is hit by `test_team_join`.

Ah right, thanks for digging it up, turns out `python3 -m unittest` runs the tests just fine w/o hitting bad things.

Considering that the peer is a mock in this test, which of these would be a better way to test this?

* Call `on_connect` with a side-effect tied to the mock's `eventData` field.
* Force `game_version` to `GAME_VERSION_AOS075` (or possibly iterate through a list of all valid versions).
* Allow setting `game_version` in the `ServerConnection` constructor.

In future I'd be very much in favour of having unit tests inherit from a base class where one could do, say, `self.make_new_server_connection()` and it'd spit out a suitable `ServerConnection` instance.

iamgreaser:
I've gone with the second option for making the unit test pass as it does less damage to the structure of the unit tests.

pyspades/packet.pyx
```diff
@@ -76,19 +79,24 @@ def register_packet(loader=None, server=True, client=True, extension=None):
...
-                raise KeyError(msg)
-            _client_loaders[cls.id] = cls
-
-        if server:
-            if cls.id in _server_loaders:
-                msg = ("Tried to register server packet '{!r}', but a packet "
-                       "with that id ({}) already exists.".format(cls, cls.id))
-                raise KeyError(msg)
-            _server_loaders[cls.id] = cls
+        for version in range(cls.since_version, cls.until_version, 1):
```
NotAFile: I'd prefer if instead of building the lookup table for all versions in between at initialization, a loader dictionary for a specific version was requested via parameters (and then cached).

pyspades/server.py
```diff
@@ -16,13 +16,17 @@
...
 import enet
 
+from twisted.logger import Logger
 from pyspades.protocol import BaseProtocol
 from pyspades.constants import (
-    CTF_MODE, TC_MODE, GAME_VERSION, MIN_TERRITORY_COUNT, MAX_TERRITORY_COUNT,
+    CTF_MODE, TC_MODE, MIN_TERRITORY_COUNT, MAX_TERRITORY_COUNT,
     UPDATE_FREQUENCY, UPDATE_FPS, NETWORK_FPS)
+from pyspades.constants import GAME_PROTOCOL_VERSIONS
+from pyspades.constants import GAME_VERSION_TO_STRING
```
NotAFile: imports again

pyspades/server.py
```diff
@@ -240,12 +250,19 @@ def update_network(self):
...
         # we only want to send as many items of the player list as needed, so
         # we slice it off at the highest player id
         world_update.items = items[:highest_player_id+1]
         self.send_contained(world_update, unsequenced=True)
 
+        # 0.76RC10+ compact world update
+        world_update_076 = loaders.WorldUpdate076()
+        world_update_076.items = items[:highest_player_id+1]
+        self.send_contained(world_update_076, unsequenced=True)
+
```
NotAFile: It seems unclean to construct the packets for all versions versions here.

pyspades/server.py
```diff
@@ -332,17 +349,35 @@ def get_random_location(self, force_land=True, zone=(0, 0, 512, 512)):
...
+                    connection.addCallbacks(
+                        self.got_master_connection,
+                        self.master_disconnected)
 
     def got_master_connection(self, connection):
-        self.master_connection = connection
+        log.info('Got master connection for version {!r}'.format(
+            GAME_VERSION_TO_STRING[connection.game_version],
+        ))
+        self.master_connections[connection.game_version] = connection
```
NotAFile: This could probably get a lot cleaner using `async/await`.

The restriction on only differentiating master connections only by version has obviously not aged that well.

pyspades/weapon.py
```diff
@@ -101,14 +138,47 @@ def get_ammo(self, no_max: bool = False) -> int:
...
+    @abstractmethod
+    def get_damage(self, value: int, position1: Any, position2: Any) -> int:
+        """
+        Returns the amount of damage dealt by this weapon.
+
+        value: An index of the body part we're hitting.
+        position1: The starting position of the shot.
+        position2: The position of the target we're hitting.
+        """
+        raise NotImplementedError()
```
NotAFile: Abstract methods on ABCs that are not overriden raise automatically on class creation.

PR #557, opened 2019-12-02
Change Documentation Layout, add a Linux example

NotAFile:
I've split the docs into admin and developer docs, and added an example setup  using systemd for service management.

### DIFF
--- a/doc/admins.rst
+++ b/doc/admins.rst
@@ -13,3 +13,4 @@ Piqueserver.
    included-scripts
    command-line
    docker
+   systemd-install-setup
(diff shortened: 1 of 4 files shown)


NotAFile:
@JohnDowson from https://www.freedesktop.org/software/systemd/man/systemd.exec.html#User=

> When used in conjunction with DynamicUser= the user/group name specified is dynamically allocated at the time the service is started, and released at the time the service is stopped

NotAFile:
I'll merge this since there's been no objections

Issue #561, opened 2019-12-09
Incomplete doc generation


1AmYF:
Looking at https://docs.piqueserver.org/en/latest/piqueserver.game_modes.html, it seems that some parts were omitted. Like the doc from the [infiltration script](https://github.com/piqueserver/piqueserver/blob/db54d1bf4af8b4adf61e1fb8a6945331a68fdfe8/piqueserver/game_modes/infiltration.py), the push mode options also don't show up. Couldn't see anything unusual in the underlying [rst file](https://github.com/piqueserver/piqueserver/blob/f9468d4a402f60a41b4672ad456a33dc23b1f946/doc/piqueserver.game_modes.rst) so far.

NotAFile:
It's probably a good idea to change this to be on a separate page that just shows the docstrings anyway

NotAFile:
This is now fixed

PR #563, opened 2019-12-10
Fix game_mode module doc generation

godwhoa:
* fixed import error in infilration
* renamed babel_script -> babel in rst

Closes: #561


NotAFile:
Oh, damn, I should have looked... I just made a huge change fixing this too. Sorry :(

I pushed it directly to master since it's just docs and we can always just fix them without any real consequences.

godwhoa:
@NotAFile np np :+1: 
I will close this

PR #565, opened 2019-12-14
Initial porting

iamgreaser:
This will at least get things working.

Here's what's been ported so far:

* bugfix: Bans a few exploits. The long name exploit has been patched in piqueserver so it's not in this script port, but the line draw exploit doesn't appear to be from what I can see, so I've ported over *that*.
* infinite_everything: Covers infiblocks and infiniclip. Merges them into the same thing.

Here's what we don't need to port:

* detectclient: `/client` is built in.
* teamchat: Functionality is built in.

Which leaves these:

* analyzer
* banhammer
* bots2
* db
* dw
* floor
* improvedweapons
* intelrpg2
* ircpublish
* irctools
* platform
* sculpt
* thunder

Some will very much need to go in (banhammer, bots2), while some are quite confused as to what they're trying to achieve so we'd be best to cherry-pick through and see what we want.

But on a lighter note, here's what we didn't have to port:

* afk
* aimbot2
* airstrike2
* blockinfo
* map_extensions
* memcheck
* paint
* passreload
* protect
* rapid
* ratio
* rollback
* spectatorcontrol
* squad
* trusted
* votekick
* welcome

However we may need to apply some patches to them (e.g. paint), which I'd prefer to do in separate scripts.


iamgreaser:
Ah crap please delete this PR if possible this is for a specific server only

NotAFile:
Doesn't look like you can delete PRs

PR #566, opened 2019-12-20
Improve doc for Push

1AmYF:


piqueserver/game_modes/push.py
```diff
@@ -161,24 +166,31 @@ def reset_intel_position(protocol, team):
...
-        if (connection.last_spawn_time is None or connection.last_spawn_time +
-                RESPAWN_CMD_DELAY.get() <= get_now_in_secs()):
-            if has_flag(connection):
-                connection.drop_flag()
-            connection.spawn()
-        else:
-            connection.send_chat(
-                "Please wait %s seconds before using this command again."
-                % (connection.last_spawn_time + RESPAWN_CMD_DELAY.get() - get_now_in_secs()))
+if ALLOW_RESPAWN_COMMAND.get():
```
NotAFile: why was this change made?

The old code allowed enabling/disabling the command at runtime. The new one does not.

1AmYF: Oh, wasn't aware of that. The reason for that change was that the command would still appear in the list of commands or with "/help r" while being disabled. Nevermind then though, should i redo this PR without that commit?

NotAFile: I think it's probably useful to a more useful "respawning is disabled" message or similar.

Or perhaps we could add some better way to disable commands. For example, using the permission system. I should re-visit #475 for that.

PR #567, opened 2019-12-21
WIP Preliminary implementation for setting options from environment vars

### LINKED ISSUE #433: Config via env vars
most modern services support some level of configuration via env vars, as they are an easy way to give some isolated context to processes. It would be neat if pique could map config options to env vars somehow.

Env vars to set the config dir and data folder would also be especially useful.

JohnDowson:
Resolves #433

### DIFF
--- a/piqueserver/config.py
+++ b/piqueserver/config.py
@@ -72,8 +72,8 @@ class ConfigStore():
         self._sections = {}
 
         # these are for config that isn't ever stored in the config file (yet)
-        self.config_dir = DEFAULT_CONFIG_DIR
-        self.config_file = os.path.join(DEFAULT_CONFIG_DIR, "config.toml")
+        self.config_dir = os.environ.get('PIQUE_CONFIG_DIR') or DEFAULT_CONFIG_DIR
+        self.config_file = os.path.join(self.config_dir, "config.toml")
 
     def _validate_all(self):
         for option in self._options.values():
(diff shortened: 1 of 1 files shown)


NotAFile:
Very nice, I'm looking forward to this.

One thing I'd like to mention about this approach is that it is accessing global state through environment variables. This make it very difficult to test, debug and generally reason about.

I would prefer explicitly passing the `os.environ` pseudo-dict to the config object, similar to how it's done for `update_from_dict(...)` and `update_from_file(...)`. A third function could be created, called `update_from_environment(os.environ)`. This could then be easily tested by just passing in a real dict instead of `os.environ`.

JohnDowson:
While messing with global state is bad, I don't exactly see how to achieve the behaviour of env vars always overwriting the config with explicit passing of env vars. You could always follow loading config file with updating from env, but that wouldn't feel particularly nice either.

NotAFile:
> You could always follow loading config file with updating from env

Yep, that's exactly what I was considering. I'm not sure how you could do it much cleaner. Perhaps by having a `ChainMap` of config sources... but that's a lot of effort to add.

JohnDowson:
I really don't see much of a problem in just silently overwriting settings from env, if that fact is properly documented.

NotAFile:
I think that silently overwriting is the way to go in terms of functionality. That's how most applications do it. What I mean here is how that's implemented in code. There I see loading from `os.environ` in a second step as clearly superior, because it can be easily verified to be correct in tests.

JohnDowson:
I think here ease of testing and cleanliness of code go in opposite direction.
I guess you could shadow os.environ for testing?

PR #568, opened 2019-12-22
WIP Multiple master server connections

JohnDowson:
Naive implementation, probably won't work.

### DIFF
--- a/pyspades/master.py
+++ b/pyspades/master.py
@@ -31,7 +31,8 @@ MAX_SERVER_NAME_SIZE = 31
 MAX_MAP_NAME_SIZE = 20
 MAX_GAME_MODE_SIZE = 7
 
-HOST = 'master.buildandshoot.com'
+BNS_MASTER = 'master.buildandshoot.com'
+HOSTS = [ BNS_MASTER ]
 
 
 class AddServer(Loader):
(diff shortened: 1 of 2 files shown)


pyspades/master.py
```diff
@@ -96,13 +97,16 @@ def on_disconnect(self):
...
 
 
 def get_master_connection(protocol):
-    defer = Deferred()
-    connection = protocol.connect(MasterConnection, HOST, PORT, MASTER_VERSION)
-    connection.server_protocol = protocol
-    connection.defer = defer
-    return defer
+    defers = []
+    for host in HOSTS:
```
NotAFile: Similarly to the last PR, There's an overall thing I'm trying to do, which is move all configuration out of the individual modules and into parameters passed in from `server.py`.

So here, I'd similarly prefer if instead of using global variables, the server name and port was instead passed in via parameters. This function would then return just one `Deferred`, and be called once per master server. This also allows us to create master connections with different protocol implementations once we implement extensions to that.

pyspades/master.py
```diff
@@ -96,13 +97,16 @@ def on_disconnect(self):
...
+        defer = Deferred()
+        connection = protocol.connect(MasterConnection, host, PORT, MASTER_VERSION)
```
NotAFile: We're running into a problem here. When creating the `enet.Host`, we need to specify the amount of Enet connections we are connecting to. We are currently allocating 34 of those, and we need to add more of those for additional masters.

lokka30:
Is there a linked issue for this? I don't understand what multiple masters is for :)

DryByte:
I don't know about a linked issue, but multiple masters means to connect to another "master/server list" at same time... So, if you connect your server to buildandshoot, you can at same time connect to another masterlist (i.e. http://checkpoint.penguins.win).

utf-4096:
Superseded by #800

PR #575, opened 2020-03-29
World update improvements

Jipok:
Now the world update cycle will take into account delays and adjust the physics update. It eliminates most microteleportations of players when the server is busy (for example, when someone downloads a map).
Also increased the number of world_update packages for spectators. Now, if you observe another player in the first person, then the picture is much nicer as it does not twitch.
I tested these changes on my babel+ server for half a year. I did not reveal any problems, the benefits are tangible.

### DIFF
--- a/pyspades/server.py
+++ b/pyspades/server.py
@@ -103,8 +103,8 @@ class ServerProtocol(BaseProtocol):
                             abs(vec[1] * 1.02) +
                             abs(vec[2] * 1.01))
 
-        self.world_time = time.time()
-        self.loop_lag = time.time()
+        self.world_time = time.monotonic()
+        self.loop_lag = time.monotonic()
 
 
     def _create_teams(self):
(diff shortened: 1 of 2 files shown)


NotAFile:
I'll have a better look at this later, but one thing I noticed is the use of `time.time()`, which returns "wall clock" time. This means it can go back or forwards in time. For use in relative calculations like these, you should always use `time.monotonic()` which is guaranteed to always increase by exactly one second for every second that the computer runs.

pyspades/server.py
```diff
@@ -249,9 +275,15 @@ def update_network(self):
         world_update = loaders.WorldUpdate()
         # we only want to send as many items of the player list as needed, so
         # we slice it off at the highest player id
-        world_update.items = items[:highest_player_id+1]
-        self.broadcast_contained(world_update, unsequenced=True)
-
+        world_update.items = items[:max(self.players)+1]
+        if for_spectators:
```
NotAFile, line 279: I'm not sure I like the idea of not being able to configure this at all. Ideally you'd be able to configure the frequency, or at least set it with one variable.

pyspades/server.py
```diff
@@ -249,9 +275,15 @@ def update_network(self):
...
+            for player in self.connections.values():
+                if (player.team is not None and player.team.spectator) or \
+                (player.world_object is None) or player.world_object.dead: # Dead players
```
NotAFile, line 282: Pretty sure this first check will send frequent world updates to people downloading the map. Which is not very helpful.

pyspades/server.py
```diff
@@ -249,9 +275,15 @@ def update_network(self):
...
+                    player.send_contained(world_update, True)
```
NotAFile, line 283: Same here, use the `keyword_argument=True` form, otherwise it's unclear what `True` means.

pyspades/server.py
```diff
@@ -215,26 +223,44 @@ def get_cp_entities(self):
         return entities
 
     def update(self):
-        self.loop_count += 1
+        lag = time.monotonic() - self.loop_lag
+        #if lag > (UPDATE_FREQUENCY*1.05):
+        #    print("MAIN LOOP UPDATE LAG: " + str(round(lag*1000)) + " ms")
```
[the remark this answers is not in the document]
Jipok: You said that twisted need to be disposed of. And the Logger is from there.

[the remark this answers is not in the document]
NotAFile: Yep. But until then, it's preferable to having commented-out debug prints in the codebase.

Jipok:
Also removed the code to increase ups for observers. I think to add it a little later with the ability to configure ups.

pyspades/server.py
```diff
@@ -66,7 +70,6 @@ class ServerProtocol(BaseProtocol):
     team1_name = 'Blue'
     team2_name = 'Green'
     spectator_name = 'Spectator'
-    loop_count = 0
```
NotAFile, line 69: I discovered that some scripts rely on this value.

pyspades/server.py
```diff
@@ -214,27 +220,49 @@ def get_cp_entities(self):
...
         return entities
 
-    def update(self):
-        self.loop_count += 1
+    async def update(self):
+        while True:
+            start_time = time.monotonic()
+            lag = start_time - self.world_time - UPDATE_FREQUENCY
+            if lag > 0.004:
+                log.debug("LAG before world update: " + str(round(lag*1000)) + " ms")
```
NotAFile, line 228:
```suggestion
                log.debug("LAG before world update: {lag:.1f}ms", lag=lag * 1000)
```

(Otherwise the string gets calculated every time, even in debug mode)

pyspades/server.py
```diff
@@ -214,27 +220,49 @@ def get_cp_entities(self):
             entities.append(flag)
         return entities
 
-    def update(self):
-        self.loop_count += 1
+    async def update(self):
+        while True:
+            start_time = time.monotonic()
+            lag = start_time - self.world_time - UPDATE_FREQUENCY
+            if lag > 0.004:
```
NotAFile, line 227: Still a magic value

pyspades/server.py
```diff
@@ -214,27 +220,49 @@ def get_cp_entities(self):
...
+                except Exception:
+                    traceback.print_exc()
+                self.world_time += UPDATE_FREQUENCY
+            # Update network
+            if time.monotonic() - self.last_network_update >= 1 / NETWORK_FPS:
+                self.last_network_update = self.world_time
+                self.update_network()
+
+            lag = time.monotonic() - start_time
+            if lag > (UPDATE_FREQUENCY * 0.7):
```
NotAFile, line 250: also a magic value with no explanation

PR #576, opened 2020-04-02
fix set_location

Jipok:
set_location_safe before:
image
set_location_safe after:
image
It seems that the error is not noticeable due to the fact that the server trusts the calculation of physics to the players.


NotAFile:
Some analysis showed that this should not impact code anywhere.

PR #578, opened 2020-04-08
Catch more errors in check_for_releases

Jipok:
```
...
File "piqueserver/server.py", line 728, in watch_for_releases
      self.new_release = await check_for_releases()
File "piqueserver/release.py", line 37, in check_for_releases
      latest_version = release["tag_name"]
builtins.KeyError: 'tag_name'
```
Sometimes it happens. There may be network problems.

### DIFF
--- a/piqueserver/release.py
+++ b/piqueserver/release.py
@@ -30,11 +30,10 @@ async def check_for_releases() -> Optional[Dict[str, Any]]:
     log.debug("checking latest version")
     try:
         release = await fetch_latest_release()
-    except IOError as e:
+        latest_version = release["tag_name"]
+    except Exception as e:
         log.warn("could not fetch latest version: {err}", err=e)
         return None
-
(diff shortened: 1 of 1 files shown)


NotAFile:
I'd personally like to know what causes this before just outright silencing it. That way lies hidden breakage.

Jipok:
Is this really important? I think, the main thing is not to give an error if the data did not come in an expected format. It’s probably worth adding parsing to the try block.

NotAFile:
Well, just catching `Exception` blanket catches all exceptions, even ones caused by programming mistakes. This makes it likely that mistakes are never discovered and things just break almost silently:

https://realpython.com/the-most-diabolical-python-antipattern/

If there is some issue that causes an invalid response, that should be checked for explicitly. I think that checking for `response.status == 200` will probably catch whatever is causing this.

PR #580, opened 2020-04-10
Consider protocol.tips may be empty

Jipok:
Fix this:
```
File "piqueserver/server.py", line 851, in send_tip
    line = self.tips[random.randrange(len(self.tips))]
File "/usr/lib/python3.5/random.py", line 195, in randrange
    raise ValueError("empty range for randrange()")
ValueError: empty range for randrange()
```

### DIFF
--- a/piqueserver/server.py
+++ b/piqueserver/server.py
@@ -446,7 +446,7 @@ class FeatureProtocol(ServerProtocol):
         """called after the map has been loaded"""
         self.update_format()
         self.tip_frequency = tip_frequency.get()
-        if self.tips is not None and self.tip_frequency > 0:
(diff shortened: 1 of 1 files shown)


piqueserver/server.py
```diff
@@ -446,7 +446,7 @@ def _post_init(self):
         """called after the map has been loaded"""
         self.update_format()
         self.tip_frequency = tip_frequency.get()
-        if self.tips is not None and self.tip_frequency > 0:
+        if self.tips is not None and self.tips and self.tip_frequency > 0:
```
NotAFile, line 449: `if self.tips` includes a none check, as `bool(None)` is `False`

PR #581, opened 2020-04-10
Introduce pathfinding

Jipok:
Work in progress. Plans:
- Improve bot path following(including building and digging)
- Fix path update loop
- Speed up A* search with can_build == True
- Set the correct constants for cost calculation in get_cost(**need help**)
- Get all the basic code for working with bots in piqueserver/bots.py
- Clean code
- Write a script demonstrating the work of bots. For example, a simple bot with a spade, following the nearest enemy

Current state:
https://streamable.com/tljwgn


Jipok:
I can not understand why build failed. Everything works fine for me.
Any comments are welcome.

godwhoa:
> I can not understand why build failed. Everything works fine for me.
> Any comments are welcome.

GCC version used by CI might be at play here.
Using more verbose initializer for `AStarFutureType` as [shown in docs](https://docs.python.org/3/c-api/typeobj.html#examples) might also help.

PR #584, opened 2020-04-13
Add on_game_start event

Jipok:
Well, we already have on_game_end. It seemed logical to me to add the opposite event.
Tested by adding bots with this event. It is good enough for this.

### DIFF
--- a/piqueserver/server.py
+++ b/piqueserver/server.py
@@ -570,6 +570,8 @@ class FeatureProtocol(ServerProtocol):
         self.set_map(self.map_info.data)
         self.set_time_limit(self.map_info.time_limit)
         self.update_format()
+        if self.initialized:
+            self.on_game_start()
 
     def set_server_name(self, name: str) -> None:
         name_option.set(name)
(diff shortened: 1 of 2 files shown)


NotAFile:
This function is named after `on_game_end`, but it doesn't actually do the opposite. I would expect every call to `on_game_start` to have a corresponding call to `on_game_end`. But this is not the case. For example, when playing two games on one map:

```
on_game_start (first map loaded)
on_game_end (first game ends)
on_game_end (second game ends)
on_game_start (second map loaded)
```

So this function doesn't actually represent the start or end of game, but the start or end of a map.

Jipok:
I just forgot that a round does not always end with a map change.

NotAFile:
I think this can be closed?

Jipok:
> I think this can be closed?

Why don't you want to add this event? In last commit I corrected the remark. Now it will look like this: 
```
on_game_start (first map loaded)
on_game_end (first game ends)
on_game_start (second game starts)
on_game_end (second game ends)
on_game_start (second map loaded)
```

Jipok:
I missed the documentation. It says: 
```
There is currently no hook for when the game starts. However, existing modes
assume it starts with the ``on_map_change`` call.

When a game ends, the game mode will call ``on_game_end`` to notify any other
scripts and possibly trigger a map change.
```
Do I need to remake the call on_game_start so that it causes it from the game modes, and not in on_game_end as now?

NotAFile:
I still like the idea of adding a hook for this but the PR is stalled, I'll convert it to an issue for now.

Issue #533, opened 2019-08-10
NoneType error on linebuild


1AmYF:
Joining a server and instantly switching to block tool and linebuilding can result in:
>[stderr#error]   File "-/piqueserver/pyspades/player.py", line 586, in on_block_line_recieved
>[stderr#error]     if not collision_3d(start_pos.x, start_pos.y, start_pos.z, x1, y1, z1,
>[stderr#error] AttributeError: 'NoneType' object has no attribute 'x'

Seems that in player.py `line_build_start_pos` is None because `on_weapon_input_recieved` was not called yet, where it receives the position data. I also managed to get the error several times in a row, not only after joining, but not entirely sure how to reproduce it. Tested with Voxlap.

PR #588, opened 2020-04-19
fix line_build_start_pos can be None

Jipok:
```
   File "pyspades/protocol.py", line 170, in update
     self.data_received(peer, event.packet)
   File "piqueserver/server.py", line 830, in data_received
     ServerProtocol.data_received(self, peer, packet)
   File "pyspades/protocol.py", line 122, in data_received
     connection.loader_received(packet)
   File "pyspades/player.py", line 153, in loader_received
     call_packet_handler(self, loader)
   File "pyspades/packet.pyx", line 131, in pyspades.packet.call_packet_handler
     handler(self, contained)


   File "pyspades/player.py", line 584, in on_block_line_recieved
     if not collision_3d(start_pos.x, start_pos.y, start_pos.z, x1, y1, z1,
 AttributeError: 'NoneType' object has no attribute 'x'
```
I don’t understand how this could happen.

UPD: Mb network delay and the player put the first one block with the right button? So the BlockLine package came before WeaponInput.

### DIFF
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -561,6 +561,8 @@ class ServerConnection(BaseConnection):
     def on_block_line_recieved(self, contained):
         if not self.hp:
             return  # dead players can't build
+        if self.line_build_start_pos is None:
+            return
 
         map_ = self.protocol.map
 


1AmYF:
This is related to #533 and propably comes from #464
IMO this doesn't fix the problem, it only prevents the exception. The player will still have his linebuild rejected.

NotAFile:
Isn't that the correct behavior? If a player starts an invalid line build, it should not complete, right?

1AmYF:
The exception is caused by a valid linebuild, not by the farbuild exploit (when i tested it in #533).

Jipok:
Could not repeat #533. I tried 7 times low and 10 times high ping. It seems to me that this is exclusively a network problem, the package WeaponInput is delayed.

Jipok:
Well, we can do this:
```
if self.line_build_start_pos is None:
    self.line_build_start_pos = self.world_object.position
```

PR #590, opened 2020-04-20
fix error and improve savemap.py

Jipok:
Fixed the error with `ensureDeferred`.
Added option `save_at_shutdown`.
Added commands to toggle `save_at_shutdown` and `load_saved_map`.
Added `save` command to save the map at the moment.

### DIFF
--- a/piqueserver/scripts/savemap.py
+++ b/piqueserver/scripts/savemap.py
@@ -75,7 +75,7 @@ def apply_script(protocol, connection, config):
                 path = get_path(self.map_info.rot_info.name)
             with open(path, 'wb') as f:
                 f.write(self.map.generate())
-            log.info("Map saved to '%s'" % path)
+            log.info("Map saved to '{path}'", path=path)
             return path
 
     return SaveMapProtocol, connection
--- a/piqueserver/scripts/savemap.py
+++ b/piqueserver/scripts/savemap.py
@@ -1,5 +1,11 @@
 """
-Saves current map on shutdown (and optionally loads it again on startup)
+Adds a /savemap command to save the current state of the map.
+You can specify any name with an argument, without it the map will
+be saved with the '.saved' suffix.
+With /rmsaved you can delete a '.saved' version of this map.
+
+Also adds two options: for automatically saving map at shutdown and 
+loading saved map instead of the original.
 
 Options
 ^^^^^^^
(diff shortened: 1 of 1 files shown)


piqueserver/scripts/savemap.py
```diff
@@ -18,32 +19,52 @@
...
         return VXLData(f)
 
 
 def apply_script(protocol, connection, config):
     class MapSaveProtocol(protocol):
+        SAVE_AT_SHUTDOWN_OPTION = False
 
         def __init__(self, *arg, **kw):
+            self.LOAD_SAVED_MAP_OPTION = savemap_config.option('load_saved_map', False).get()
+            self.SAVE_AT_SHUTDOWN_OPTION = savemap_config.option('save_at_shutdown', False).get()
```
NotAFile, line 59: We do this `.get()` thing because we want changes from reloading the config to apply without a restart. Doing this in `__init__` doesn't allow that.

piqueserver/scripts/savemap.py
```diff
@@ -18,32 +19,52 @@
...
 
-
 def load_map_from_path(path):
     with open(path, 'rb') as f:
         return VXLData(f)
 
 
 def apply_script(protocol, connection, config):
     class MapSaveProtocol(protocol):
+        SAVE_AT_SHUTDOWN_OPTION = False
```
NotAFile, line 55: `ALL_CAPS`, in python convention, indicates a constant. This is not one, since it is modified.

piqueserver/scripts/savemap.py
```diff
@@ -18,32 +19,52 @@
...
 from pyspades.vxl import VXLData
 from piqueserver.config import config
+from piqueserver.commands import command
 
+savemap_config = config.section('savemap')
 log = Logger()
 
-savemap_config = config.section('savemap')
-LOAD_SAVED_MAP_OPTION = savemap_config.option('load_saved_map', False)
+@command('save', admin_only=True)
```
NotAFile, line 27: Save can mean lots of things, i think this should be called `/savemap`

piqueserver/scripts/savemap.py
```diff
@@ -18,32 +19,52 @@
...
+    protocol = connection.protocol
+    protocol.SAVE_AT_SHUTDOWN_OPTION = not protocol.SAVE_AT_SHUTDOWN_OPTION
+    log.info("SAVE_AT_SHUTDOWN: %s" % protocol.SAVE_AT_SHUTDOWN_OPTION)
 
 
 def get_name(map_info):
     return '%s/%s.saved.vxl' % (os.path.join(config.config_dir, 'maps'),
                                 map_info.rot_info.name)
 
-
```
NotAFile, line 32: pep8 violation (we should really run an automatic checker for this...)

piqueserver/scripts/savemap.py
```diff
@@ -18,32 +19,52 @@
...
 log = Logger()
 
-savemap_config = config.section('savemap')
-LOAD_SAVED_MAP_OPTION = savemap_config.option('load_saved_map', False)
+@command('save', admin_only=True)
+def save(connection):
+    connection.protocol.save_map()
+
+@command('toggle_load_saved_map', admin_only=True)
+def toggle_load_saved_map(connection):
```
NotAFile, line 32: I'm not convinced this is something you'd want to turn on and off regularly enough to justify an in-game command. Especially since the server has already started and loaded a map at this point.

piqueserver/scripts/savemap.py
```diff
@@ -18,32 +19,52 @@
...
+    protocol = connection.protocol
+    protocol.LOAD_SAVED_MAP_OPTION = not protocol.LOAD_SAVED_MAP_OPTION
+    log.info("LOAD_SAVED_MAP: %s" % protocol.LOAD_SAVED_MAP_OPTION)
```
NotAFile, line 35: This is wouldn't be a useful log message, especially for `info` level. Players using it also don't get any feedback what they set the value to because it doesn't return anything.

piqueserver/scripts/savemap.py
```diff
@@ -18,32 +19,52 @@
...
+
+@command('toggle_save_at_shutdown', admin_only=True)
+def toggle_save_at_shutdown(connection):
```
NotAFile, line 38: Same for this. What scenario do you have in mind that you'd want to regularly toggle this from in-game?

piqueserver/scripts/savemap.py
```diff
@@ -8,67 +14,68 @@
...
+
+        def save_map(self, custom_name=None):
+            if custom_name:
+                path = os.path.join(config_dir, 'maps', custom_name) + '.vxl'
+            else:
+                path = get_path(self.map_info.rot_info.name)
+            with open(path, 'wb') as f:
                 f.write(self.map.generate())
-            log.info("Map saved to {}".format(get_name(self.map_info)))
+            log.info("Map saved to '%s'" % path)
```
NotAFile, line 78:
```suggestion
            log.info("Map saved to '{path}'", path=path)
```

Jipok: And for what?

NotAFile: This is just the correct way of using the logging api. The formatting code won't be run unless the message is actually printed and you can emit the logs as easily parseable JSON if you want.

PR #593, opened 2020-04-22
Fix of a player name

Rakete175:
Invisible players should be addressed with their name instead of the name of the person they are shooting


NotAFile:
Thanks for the PR. However, this is working as intended. The intent of the message isn't to address the player, but to allow you to discover the names of players on the other team. Perhaps the message used could be clearer though.

PR #595, opened 2020-04-24
Babel game mode bug fix

Samuel-Cavicchi:
The ability to shoot and destroy blocks was previously based on the player's position. This has been updated to check based on the block's position.


NotAFile:
This is working as intended. The intent of this behavior is to prevent shooting the tower from your side of the map.

We can absolutely have a discussion about ways in which the destruction rules could be improved though. Feel free to open issues for problems with this behavior, or discuss them in chat.

PR #596, opened 2020-04-30
increase can_see distance limit to 128

Jipok:
This function is only used in nointelonwalls and nospadingwalls scripts. That is, the distance does not really exceed 32 blocks. In fact, even less.
And also used in calculating grenade damage. Which never exceeds 16 blocks.
Therefore, this fix does not change anything for the current codebase. However, now the name is true. And it will help my script against esp a lot.

### DIFF
--- a/pyspades/world_c.cpp
+++ b/pyspades/world_c.cpp
@@ -237,8 +237,8 @@ long can_see(MapData *map, float x0, float y0, float z0, float x1, float y1,
     ftol(f.y * g.x - f.x * g.y, &p.z);
     ftol(g.z, &i.z);
 
-    if (cnt > 32)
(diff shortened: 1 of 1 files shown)


NotAFile:
I'm curious what the performance impact of this is.

Jipok:
You can check yourself, `can_see` is used in two scripts: `nointelonwalls.py` and `nospadingwalls.py`
In both cases, the verification distance does not exceed the pair of blocks. So even an old limitation is not achieved. 

But even if you imagine that it was achieved, then the check is very, very fast, and the dependence on the number of blocks is linear

Briefly speaking, 0 performance impact

NotAFile:
Sounds good

PR #609, opened 2020-11-25
Bugfix of inability to load from config - onectf.py

EvgenZhaba:
**Describe the bug**
 In the script "piqueserver.game_modes.onectf" you cannot load the selected mode from the config. The default is REVERSE_ONE_CTF.

**To Reproduce**
1. select game mode "piqueserver.game_modes.onectf"
2. add the line "extensions = { 'one_ctf': True }" to the txt file for the map
3. run piqueserver
4. take intel and see the message REVERSE_ONE_CTF_MESSAGE = 'Take the intel to the enemy base to score.'

**Additional context**
https://github.com/piqueserver/piqueserver/blob/master/piqueserver/game_modes/onectf.py#L18 
ONE_CTF_MODE = REVERSE_ONE_CTF

https://github.com/piqueserver/piqueserver/blob/master/piqueserver/game_modes/onectf.py#L58
will always exit here without reading config

### DIFF
--- a/piqueserver/game_modes/onectf.py
+++ b/piqueserver/game_modes/onectf.py
@@ -53,14 +53,14 @@ def apply_script(protocol, connection, config):
             self.one_ctf = self.reverse_one_ctf = False
             self.one_ctf_spawn_pos = FLAG_SPAWN_POS
             extensions = self.map_info.extensions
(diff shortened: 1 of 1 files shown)


NotAFile:
Thanks a lot! Sorry it took me so long.

PR #613, opened 2021-01-05
badmin.py: Check also for rifle percentage

Haxk20:
Without this change whenever you would check for a player that is using rifle with /investigate
the command would fail because it did not return anything thus being type None which would
cause issues with round().

semi_hits doesn't have to exist as we have rifle_hits which is what we should use for checking 
rifle hits and counts.

With this fix badmin finally checks also for players using rifle and doesn't fail on that.

### DIFF
--- a/piqueserver/scripts/badmin.py
+++ b/piqueserver/scripts/badmin.py
@@ -192,16 +192,12 @@ def score_grief(connection, player, time=None):  # 302 = blue (0), #303 = green 
 
 
 def check_percent(self):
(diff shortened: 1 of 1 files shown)


Haxk20:
@NotAFile Could you please review this and possibly merge it when you have time ?

NotAFile:
@Haxk20 thanks, sorry it took me so long!

Haxk20:
> @Haxk20 thanks, sorry it took me so long!

No problem. Just glad it got merged :)

PR #615, opened 2021-02-28
@player_only fix and airstrike2 update

DeletedUser92:
- Fix a bug in @player_only, where the command stop working fine or dont work.
- Added support to do /airstrike <COORD> (like /airstrike B4) for use the airstrike.

### DIFF
--- a/piqueserver/commands.py
+++ b/piqueserver/commands.py
@@ -265,7 +265,7 @@ def player_only(func: Callable):
     def _decorated(connection, *args, **kwargs):
         if connection not in connection.protocol.players.values():
             raise CommandError("only players can use this command")
-        func(connection, *args, **kwargs)
+        return func(connection, *args, **kwargs)
     return _decorated
 
 def target_player(func: Callable):
--- a/piqueserver/scripts/airstrike2.py
+++ b/piqueserver/scripts/airstrike2.py
@@ -14,7 +14,7 @@ from math import ceil, sin, cos
 from random import uniform, vonmisesvariate
 from twisted.internet import reactor
 from pyspades.contained import GrenadePacket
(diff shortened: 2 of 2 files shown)


NotAFile:
Hi, thank you for your pull-request.

However unfortunately I don't think this should be merged for two reasons:

 - The PR and commit combines a bugfix with an unrelated change. These should be separate.
 - Needing to see the target of the airstrike is deliberate game design of the airstrike2 script. The radius of the `airstrike2` strike is also significantly smaller than one sector, so having it as a command does not make a lot of sense.
 
I am open to adding a new `classic_airstrike` script (or config option) that replicates the classic airstrike behavior of striking a whole map sector, but that requires more changes than what's in this PR.

I've applied your bugfix in 269356a and welcome further discussion of how to bring classic airstrikes to pique.

lokka30:
> Hi, thank you for your pull-request.
> 
> However unfortunately I don't think this should be merged for two reasons:
> 
> * The PR and commit combines a bugfix with an unrelated change. These should be separate.
> * Needing to see the target of the airstrike is deliberate game design of the airstrike2 script. The radius of the `airstrike2` strike is also significantly smaller than one sector, so having it as a command does not make a lot of sense.
> 
> I am open to adding a new `classic_airstrike` script (or config option) that replicates the classic airstrike behavior of striking a whole map sector, but that requires more changes than what's in this PR.
> 
> I've applied your bugfix in [269356a](https://github.com/piqueserver/piqueserver/commit/269356a20c5858c7290f0a43ce5a1f319efe8647) and welcome further discussion of how to bring classic airstrikes to pique.

`classic_airstrike` would be awesome!

PR #621, opened 2021-04-22
Fix votekick.py

Rakete175:
This PR fixes the most important bugs of the votekick script.

The main thing this PR changes is that votekicks are now IP dependent.

Although the coding style may not be 100% perfect, I strongly recommend merging a fix like this.

### DIFF
--- a/piqueserver/scripts/votekick.py
+++ b/piqueserver/scripts/votekick.py
@@ -182,7 +182,10 @@ class Votekick:
     def start(cls, instigator, victim, reason=None):
         protocol = instigator.protocol
         last_votekick = instigator.last_votekick
-        reason = reason.strip() if reason else None
+        reason = reason.strip() if reason else None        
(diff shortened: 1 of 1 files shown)


NotAFile:
Hm, I'm really not quite sure what this does. There seem to be a lot of unrelated changes in one commit and it's not really clear why they are necessary. Do these relate to any specific issues?

I can tell that it's preventing players with the same IP address from voting multiple times, but there's many legitimate reasons for players to share an IP address. Pique also already limits the number of connections per IP address. Have you encountered any specific cases where this would have been helpful?

We also follow the pep8 style: https://www.python.org/dev/peps/pep-0008/

NotAFile:
After some discussion, we've decided to rework this patch into a new PR

NotAFile:
Reopening because this would still be useful and I don't remember why we closed it initially.

PR #628, opened 2021-05-20
Skip on_world_update on heavy lag

Jipok:
This is an important part of my past pr with a change in world_update. I do not know why I did not send this commit. Maybe i wanted to test and pick up best number than 2 seconds.
It seems that it was discussed in chat a year ago. I do not remember. 

This change avoids the stuck in the world update cycle. When the scripts are heavy (or poorly written), then on_world_update can take longer than the budget allows. Accordingly, the code will try to catch up and make another Update without waiting and update_network. Thus, it can never get out of the cycle or freeze for a long time(minutes).

I tested with bots. If I launched 30 bots and made the calculation of the path every tick, then without this commit, the server periodically hangs on the time from 30 seconds to 2 minutes. What is very bad for players. 
With this patch, the freezing is not noticeable. Just a small lag for a player, even the downloading of the map is relatively smooth.


pyspades/server.py
```diff
@@ -236,13 +236,16 @@ async def update(self):
...
             # Update world
-            while (time.monotonic() - self.world_time) > UPDATE_FREQUENCY:
+            while (time.monotonic() - self.world_time) >= UPDATE_FREQUENCY:
                 self.loop_count += 1
                 self.world.update(UPDATE_FREQUENCY)
-                try:
-                    self.on_world_update()
-                except Exception:
-                    traceback.print_exc()
+                if (time.monotonic() - start_time) < 2:
```
NotAFile, line 242: I'd want this to be a constant, something like `UPDATE_DEFICIT_MAX`. It would also be good to only get the value of `time.monotonic()` once so that is does not change during the update. That can cause weird behavior.

pyspades/server.py
```diff
@@ -236,13 +236,16 @@ async def update(self):
...
+                    try:
+                        self.on_world_update()
+                    except Exception:
+                        traceback.print_exc()
+                else:
+                    log.warn("world update heavy LAG. on_world_update skipped")
```
NotAFile, line 248: I'd say something like "world update behind by %.2s seconds, skipping update".

NotAFile (changes requested):
In general, I am in favor of handing this case, the spiral of death. But I'm not sure about the way it's handled here.

Generally there's two ways of handing it:
 - limit the amount of loops per update, effectively slowing down the simulation
 - discard the whole backlog and reset to a clean state

This does a weird half-thing where it just skips single updates at a time. I'd suggest just fully giving up instead and setting `self.world_time` to the current time. Then teleport all players to their current server positions to make sure everything is synced.

Jipok:
All your options involve skipping the physics calculation. Which causes eerily annoying back-teleportations for clients.
I think the physics calculation is important. A bad script can be skipped. I have never seen scripts that break because of this.

pyspades/server.py
Jipok: > It would also be good to only get the value of time.monotonic() once so that is does not change during the update.

But it should change during the update. This is the point.

NotAFile:
> All your options involve skipping the physics calculation. Which causes eerily annoying back-teleportations for clients.

That is inevitable either way though. If we are two seconds behind on updates, that means that the server has missed two seconds of player activity. There's nothing it can really do to catch up on that. I think it is then better to just accept the teleportation and make sure that we have a consistent state for the future instead of doing unpredictable half-updates. If you are regularly behind for more than two seconds you have much bigger problems anyway.

Jipok:
Hmm. Now I understand that this is incorrect. We may have a delay due to something else, but on_world_update will be skipped.
It is necessary to take into account the execution time of exactly on_world_update.

PR #629, opened 2021-06-03
Rewrite update_network. 60 ups for specs, 10 for players. New event on_network_update

Jipok:


Jipok:
It can be merged as it stands. 
In an amicable way, it would be necessary to define the spectator in a different code. But then, in an amicable way, many other things must be rewritten. Unfortunately, I have no desire for this.

pyspades/server.py
```diff
@@ -282,7 +285,42 @@ def update_network(self):
...
+                continue
+            spec = False
+            # 60 ups only for spectators
+            if player.team is not None and player.team.spectator:
+                spec = True
+            # New players without a team are spectators too 
+            if player.team is None and player.saved_loaders is None:
+                spec = True
+            # Dead players are spectators too 
+            if player.world_object is not None and player.world_object.dead:
```
Haxk20, line 307: While this makes sense from logical standpoint i dont think its great idea to switch player UPS midgame like this. Mainly because users with very bad connections will struggle with this.  So when they die their UPS will go to 60 and then back to 10 UPS.

In my opinion dead players should just have 10UPS.

Haxk20 (changes requested):

pyspades/server.py
Jipok: Determine very bad connection conditions. I tested with varying conditions on the brink of playability. Increasing ups has no effect at all. As in and should not in fact.
In any case, I decided to fork the project with my many changes. Very slowly, but I'm moving forward. I will publish someday.

Haxk20: Bad conditions are the fact that when we forced TPS to 30 on AOSG some players were struggling with connections and lag on their side occurred. The same occurred on aloha when we tried to switch to 20 TPS. In the later we even saw players getting out right disconnected. Their connections just could not handle the few more KB/s. I learned the hard way that i should not underestimate how bad some players connections are.

Jipok: Well this is interesting. How you implement raising ups? My version has been hosted for almost a month. Players enter, there have never been any complaints or incomprehensible disconnections.
I also know a little about how it works at the network level. Network load of 60 ups is ridiculous compared to some modern bloat website.

Haxk20: UPS cant even be compared to a modern website due to how it works on the network level.

Website has a LOT of time to load. Better say its not time limited to finish loading.
Tho UPS on the other side is time limited. It must arrive at the client faster or at 16ms when on 60UPS. Otherwise the UPS drops and it will feel like frame drops even tho your FPS may be pinned at 300fps.

And we cant forget about other packets that are sent to clients.

And its not all about the network usage. If the ping is extremely unstable and very high (300ms and jumping from 300-500ms) then players will see a much higher network usage then what you are sending to them due to how late the packets arrive on their side. On server side you may be sending the UPS at 16ms spacing for 60UPS but on their side it may arrive depending on ping in different intervals. Sometimes even 5ms between them. 
This is what is causing the issue i have told about. Its the unstability. We do not see those issues as we send UPS normally at 100ms and there is a time to not get the packets packed into a tiny time window with lots of packets which can happen if you send UPS way faster.

And while this is only occuring for a very small amount of players you cant exactly make the game unplayable for them.
Even if the issue was occuring for 1 player the AoS player base is small enough already that ever making it unplayable for 1 player is not great.

Thats why i recommend leaving the ups when player dies at 10 instead of switching it to 60 for the tiny duration of time.

PR #707, opened 2021-12-09
Check distance in hitpacket handler

MuffinTastic:
... and silently reject any hits from well beyond fog range. This prevents cheaters from making literally impossible shots from the whole map away. We've had this running on aloha for ages without issue.

Doing the sqrt() in plain view personally feels less dirty to me than creating a Vertex3 just to use 2/3 of its fields and dispose of it immediately, but if you guys would rather me do that I'll change it, no big deal.
Also, I'm slightly annoyed at pyspades' lack of a Vector2/Vertex2 class.

It's a mystery to me as to why FOG_DISTANCE doesn't normally exist in the first place.

### DIFF
--- a/piqueserver/scripts/analyze.py
+++ b/piqueserver/scripts/analyze.py
@@ -61,7 +61,7 @@ def apply_script(protocol, connection, config):
         prev_time = None
         body_part = ""
 
-        def on_hit(self, hit_amount, hit_player, type, grenade):
+        def on_unvalidated_hit(self, hit_amount, hit_player, type, grenade):
             if self.name in list(self.protocol.analyzers.values()):
                 if type == HEADSHOT_KILL or hit_amount in body_damage_values or hit_amount in limb_damage_values:
                     if not grenade:
@@ -101,7 +101,7 @@ def apply_script(protocol, connection, config):
                                     else:
                                         analyzer.send_chat('%s shot %s dist: %d blocks dT: NA %s %s(%d)' % (
                                             self.name, hit_player.name, dist, weap, body_part, counter))
-            return connection.on_hit(self, hit_amount, hit_player, type, grenade)
+            return connection.on_unvalidated_hit(self, hit_amount, hit_player, type, grenade)
 
         def on_weapon_set(self, value):
             if self.name in list(self.protocol.analyzers.values()):
--- a/pyspades/constants.py
+++ b/pyspades/constants.py
@@ -44,6 +44,8 @@ SPAWN_RADIUS = 32
 MINE_RANGE = 3
 BUILD_TOLERANCE = 5
 
+FOG_DISTANCE = 128.0
+
 MELEE_DISTANCE = 3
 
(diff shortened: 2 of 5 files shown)


Rakete175:
Definitely a good idea, especially because fogshots are easy to detect.

However, I would suggest to move that code below the called on_hit function. In your current way this is a handicap against people who just want to catch cheaters with analyze.py, because those fogshot attempts won't be visible there.

Haxk20:
Bravo.

MuffinTastic:
> However, I would suggest to move that code below the called on_hit function. In your current way this is a handicap against people who just want to catch cheaters with analyze.py, because those fogshot attempts won't be visible there.
Hmm, I'm torn on this. I personally feel on_hit should really only be for valid hits, as it's already blocked from being called by melee hitpackets that are out of range... and if you really want the invalid packets, you can do some hacky workarounds. But you're probably right, so I'll change it.

NotAFile:
I could swear this check was already done elsewhere? Are you sure this is not the case?

Haxk20:
> I could swear this check was already done elsewhere? Are you sure this is not the case?

*cough* *cough* *cough* SpadesX

MuffinTastic:
I searched the codebase for any reference to `hit`, to see what might hook off of the connection class's hit and on_hit methods. There are a few scripts that use it (namely analyze) but I came up with nothing that actually stops the hitpacket. I also looked at world_object/Character's validate_hit, it doesn't do a distance check either. I don't think it's in there, but feel free to check from your end, one can never be too sure.

NotAFile:
Yeah you might be right there @MuffinTastic . It seems I got confused with the checks performed in analyze and the checks performed for other actions by can_see.
Might be a good idea to put this into validate_hit instead to keep things together.

Haxk20 (changes requested):
Squash it into a single commit locally and force push :)

NotAFile:
I can still squash things when merging and it's not bad to preserve the history

MuffinTastic:
> Might be a good idea to put this into validate_hit instead to keep things together.

I agree, however, doing so would require either new hacky workarounds within scripts like analyze or a new hook (some `on_unvalidated_hit`) for them to use, as @Rakete175 pointed out that on_hit is currently used in situations where the hitpacket should really be invalid, for better or worse. If that's the way we want to go I'll do it.

Well, either it's that or we put the call to validate_hit after the call to on_hit, which to me seems rather... backwards.

NotAFile:
That is a good point. However I feel like more scripts would break from letting invalid hits through, wanting to receive invalid hits seems like a very rare case. It would be annoying to have to add these checks to all of them. Perhaps an event for invalid hits would be ideal for the `analyze` case?

Jipok:
Constant 128 is also used in the part of the code written in c/c++: `world_c.cpp -> can_see`
I just point out the fact. This is hardly why it is worth transferring the check from python to C.

NotAFile:
looks good to me!

PR #709, opened 2021-12-09
Filter some malicious characters from names and chat

MuffinTastic:
For a short while '#' was used a lot by some people on our servers to avoid getting votekicked, as it interfered with player IDs.

### DIFF
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -606,6 +606,7 @@ class ServerConnection(BaseConnection):
                 return
             elif result is not None:
                 value = result
+            value = value.replace('\n', '')
             contained.chat_type = CHAT_ALL if global_message else CHAT_TEAM
             contained.value = value
             contained.player_id = self.player_id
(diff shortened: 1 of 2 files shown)


NotAFile:
IIRC, we reworked the votekick script a long time ago to prevent this exploit

NotAFile:
Yeah, votekick uses get_player which we reworked to prevent this issue:
https://github.com/piqueserver/piqueserver/blob/master/piqueserver/commands.py#L310

I don't object to the removal of newlines, although we also do some cleanup elsewhere.

Jipok:
The cluttered (and empty) names are really annoying. Maybe should filter based on [Unicode categories](https://stackoverflow.com/a/93557)?

NotAFile:
I think this has been superseded by #804

utf-4096:
#804 only applies to chat messages, not nicknames. I'll make a new PR that applies the current filters to nicknames in a moment.

PR #710, opened 2022-01-01
Update airstrike2.py

Eneroh:
/a will now work when inputting command into chat


NotAFile:
The problem here was that it would not show the message if an airstrike was available, right?

NotAFile:
Actually, I don't see the problem here. What does this fix?

Eneroh:
When you type /a it didn't count how many kills until you get until an airstrike. Minor fix. I tested the script with my server and it works so eh.

NotAFile:
Oh, I see. That makes sense.

PR #721, opened 2022-06-09
Add command anti-spam

lokka30:
This PR adds a basic anti-spam system for commands.

Currently, commands can be spammed by users which bakes the Python interpreter. This is especially prevalent with commands that utilize `send_lines`, for which I will address in a separate pull request. I hope that this addition will make it more difficult for commands to be exploited by malicious users.  

- [x] Add command anti-spam
- [x] Comment on initial commit from maintainer
- [x] Refactor to use existing rateliit tools
- [x] Add toggle in config
- [x] Allow for configurable delay
- [ ] Test
- [ ] Ready for merge

If there are any changes required for this pull request, please let me know.


NotAFile:
Hey, thanks for the fix. I think the approach I would like to see instead is to use the existing rate limiting tools that are used for other things too.

lokka30:
Cheers nota. I'll give that a shot 🙂

lokka30:
@DryByte may also work on this PR. Apologies for my inactivity. 😄

DryByte:
Im not sure about the config implementation... Can someone check if its fine in that way?

piqueserver/config/config.toml
```diff
@@ -168,6 +168,12 @@ time_announcements = [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 30, 60, 120, 180, 240, 300
 # number of intel captures to win the game (default: 10)
 cap_limit = 10
 
+# Enable command antispam for preventing users to abuse from flooding commands
+enable_command_antispam = true
```
lokka30, line 172: It may be out-of-scope for this PR, though having a common configuration section for antispam features might make the config more consistent.

lokka30: I believe they're also called 'ratelimits' internally - so we may also want to rename this to 'enable command rate-limiter' or something along those lines. If we do that then also make sure to rename the other settings.

piqueserver/config/config.toml
```diff
@@ -168,6 +168,12 @@ time_announcements = [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 30, 60, 120, 180, 240, 300
...
+
+# Maximum commands to be executed in the given time
+max_command_size = 4
```
lokka30, line 175: Without reading the comment supplied with the setting, it may seem like 'max command size = how long a command can be; only 4 characters'. Perhaps adding a plural (s character) would help: `command_ratelimit_amount`.

piqueserver/config/config.toml
```diff
@@ -168,6 +168,12 @@ time_announcements = [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 30, 60, 120, 180, 240, 300
...
+max_command_per_time = "5s"
```
lokka30, line 176: Perhaps a better name for this setting is `command_ratelimit_period`.

piqueserver/server.py
```diff
@@ -461,10 +471,11 @@ def _next_time_announce(self):
             if remaining < 10.001:
                 self.broadcast_chat('%s...' % int(round(remaining)))
             else:
-                self.broadcast_chat('%s seconds remaining.' % int(round(remaining)))
+                self.broadcast_chat('%s seconds remaining.' %
+                                    int(round(remaining)))
         else:
             self.broadcast_chat('%s minutes remaining.' %
-                           int(round(remaining / 60)))
+                                int(round(remaining / 60)))
```
lokka30, line 478: Some of these formatting changes are out-of-scope for this PR, assuming they were done automatically by your IDE

DryByte: These changes are actually made by autopep8 that tries to keep the formatting standards in pep8. Thats why i made it in a separete commit, but we can revert it anyway, i just kept it because it will need to get updated at some point

DryByte:
@lokka30 can you check if its fine for you?

lokka30 (commented):
LGTM!

NotAFile (approved):

Issue #624, opened 2021-04-25
Users can spam commands to destroy the CPU


lokka30:
**Describe the bug**
Users can spam commands such as `/commands` to destroy the CPU.

I believe this issue is solely caused by the 'send user a list of messages' code. See **Expected behavior** below.

To fix this, I believe the commands should be locked until the previous usage of the command has finished sending the entire list of messages.

**To Reproduce**
Have a bot rapidly run list-based commands such as `/commands`, and most likely also `/scripts`, `/help`, and so on.

**Expected behavior**
This is where the issue kind-of separates into a feature request. Here's the intended behavior I am thinking of:
1. User runs `/commands`
2. Commands start to display to the user
3. Two seconds later, whilst the previous command is still being ran, the user runs `/commands` again
4. Second command fails as the user is already being displayed the output of `/commands` from their first usage.

**Stack Trace or logs**
Well, my logs would be full of someone destroying the server by constantly running such commands.

**Server (please complete the following information):**
Not applicable to the issue.

**Client (if relevant)**
Not applicable to the issue.

**Additional context**
I have temporarily disabled these commands on The Hallway due to their misuse.

lokka30:
This is resolved by my two PRs, #721 and #722

PR #722, opened 2022-06-09
Patch `send_lines` exploit

lokka30:
# About

> Note: This should be merged alongside #721 to fully patch the two exploits mentioned in #624.

Piqueserver does not prevent users from spamming commands which utilize the `send_lines` function. For commands that send a relatively larger amount of lines in the `send_lines` function, such as `/commands`, this exacerbates the issue.

This pull request does not deal with command spam (which has been left to #624) – this pull request *does* deal with multiple `send_lines` calls of the same underlying reason.

# Explanation of implementation

To address this exploit, I have added a set "`current_send_lines_types`" to `FeatureConnection`, which stores the list of types that a player is currently being sent.

Types are simply 'reasons' that `send_lines` has been called, which serve as identifiers, so that `send_lines` can be called only once at a time for a particular message.

For example, running `/commands` uses the type `commands` for the `send_lines` function.

If `send_lines` is called but the user is already receiving the lines of the given `type`, then an `info`-level message is logged to the console to inform the server that the function call has been silently discarded.

# To-do

- [x] Initial commit to fix exploit
- [x] Maintainer's satisfaction with the direction of this PR
- [x] Document the `type` parameter on `send_lines`
- [x] Mark pull request as 'ready to merge'

### DIFF
--- a/piqueserver/player.py
+++ b/piqueserver/player.py
@@ -334,6 +334,7 @@ class FeatureConnection(ServerConnection):
                                       self.name)
 
     def send_lines(self, lines: List[str]) -> None:
+        # TODO: address here
         current_time = 0
         for line in lines:
             reactor.callLater(current_time, self.send_chat, line)
--- a/piqueserver/player.py
+++ b/piqueserver/player.py
@@ -44,7 +44,7 @@ class FeatureConnection(ServerConnection):
         self.user_types = None
         self.rights = None
         self.can_complete_line_build = True
-        self.current_send_lines_types = {}
+        self.current_send_lines_types = []
 
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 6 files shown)


piqueserver/player.py
```diff
@@ -333,12 +335,25 @@ def ban(self, reason=None, duration=None):
                 self.protocol.add_ban(self.address[0], reason, duration,
                                       self.name)
 
-    def send_lines(self, lines: List[str]) -> None:
+    def send_lines(self, lines: List[str], type: str = 'unknown') -> None:
```
NotAFile, line 338: I would rename type to key, as type is ambiguous and collides with the built-in type function

piqueserver/player.py
```diff
@@ -43,6 +44,7 @@ def __init__(self, *args, **kwargs):
         self.user_types = None
         self.rights = None
         self.can_complete_line_build = True
+        self.current_send_lines_types = {}
```
NotAFile, line 47: `{}` creates a dict, not a set, so this will not run

NotAFile (changes requested):
Overall I am happy with this approach.

There's a few other things that could be done, like only using a single task, cancelling the previous send_lines if a new one appears, but this should be good enough for now.

lokka30:
Thanks for checking it out nota. I've attempted to resolve your suggestions, along the way I've also added a comment for the function and made it skip any detection if the default/unspecified key is used.

I have a thought that it might be better to add a function with the old syntax back (only the `lines` parameter), make that deprecated, and remove the default key on the new function (so `key: str` instead of `key: str='unknown'`). The 'old' function will then just call the new one using the default key.

PR #723, opened 2022-06-12
Speed up client detection

DryByte:
**Description**
- Client detection was being slow, and when using detectclient script the detection by detectclient was being really faster than piqueserver one. When removing the packets sent by piqueserver core, detectclient is slow like piqueserver

**How to reproduce?**
- Join in the server with the benchmark script (at the end of this)

**Why this need to be merged?**
- Because some scripts need to know the user's client in the fastest way for doing some gamemode/script change.

Detect client speed when using piqueserver's _connection_ack handshake:

Benchmark before the changes:
Screenshot_2022-06-12-12-17-50_1024x768

Benchmark after the changes:
Screenshot_2022-06-12-12-20-45_1024x768

Benchmark script:
```py
from time import time
from pyspades.packet import register_packet_handler
from pyspades.contained import VersionResponse

	class ffsConnection(connection):
		def __init__(self, *args, **kw):
			self.testStart = time()
			return connection.__init__(self, *args, **kw)

		@register_packet_handler(VersionResponse)
		def on_version_info_recieved(self, contained):
			print("[PIQUESERVER] - In: {}".format(time()-self.testStart))
			return connection.on_version_info_recieved(self,contained)

		def on_version_answer(self, info):
			print("[DCLIENT] - In {}".format(time()-self.testStart))
			return connection.on_version_answer(self, info)

	return protocol, ffsConnection
```

### DIFF
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -657,7 +657,7 @@ class ServerConnection(BaseConnection):
     @register_packet_handler(loaders.HandShakeReturn)
     def on_handshake_recieved(self, contained: loaders.HandShakeReturn) -> None:
         version_request = loaders.VersionRequest()
-        self.protocol.broadcast_contained(version_request)
+        self.send_contained(version_request)
 
     @register_packet_handler(loaders.VersionResponse)
     def on_version_info_recieved(self, contained: loaders.VersionResponse) -> None:


NotAFile (approved):

NotAFile:
surprised this slipped through, thanks for the fix!

Issue #619, opened 2021-04-09
Bans seem to be disappearing


Rakete175:
**Describe the bug**
A description of what the bug is.
Bans are disappearing and already banned people can sometimes join again.

**To Reproduce**
1. Create a server
2. Ban people for permanent duration
3. (Maybe restart the server)
4. Wait until they rejoin

**Expected behavior**
A description of what you expected to happen.
Bans should persist

**Stack Trace or logs**

https://gist.github.com/Rakete175/4c7b9e3ca30413ba53422facdd04dfbb

**Server (please complete the following information):**
 - OS/Version: CentOS7
 - Piqueserver Version: 1.0.0
 - Installation Method: pip3

**Client (if relevant)**
 - Name: [e.g. OpenSpades, BetterSpades]
 - Version: [e.g. 1.3.0]

**Additional context**
The gist shows a problem which might be related to that (32780 -> 32771 bans within 18 hours)

lokka30:
We _appear_ to be having this issue as well on The Hallway, although I am uncertain. Lots of bans go in and out, so it's difficult to keep track.

NotAFile:
Interesting. I presume this is some kind of mistake in the ban vacuum code? Or perhaps a mistake in serialization.

WIth debug log level, it would be possible to see if any bans were removed by the ban vacuum.

Rakete175:
Thanks for your answer! 
Now I used the  "debug" loglevel for a few hours and when I search for the word "ban*" I get the following results:

22149:2021-04-14T05:27:43+0200 [piqueserver.server#info] starting ban vacuum with 32486 bans
22150:2021-04-14T05:27:43+0200 [piqueserver.server#debug] ban vacuum took 0.35 seconds, removed 3 bans
22151:2021-04-14T05:27:44+0200 [piqueserver.server#debug] saving 32483 bans took 0.47 seconds
32268:2021-04-14T11:27:43+0200 [piqueserver.server#info] starting ban vacuum with 32483 bans
32269:2021-04-14T11:27:43+0200 [piqueserver.server#debug] ban vacuum took 0.23 seconds, removed 3 bans
32270:2021-04-14T11:27:44+0200 [piqueserver.server#debug] saving 32480 bans took 0.43 seconds

NotAFile:
Interesting that it's removing exactly three every time

Rakete175:
Update:
I got the idea that I could log, who these bans are.
And interestingly it were always the same entries which were removed 🤔

[...]

PR #729, opened 2022-06-16
Wrong bans being deleted

DryByte:
**Description**
- This PR is related to the issue #619, bans was always deleting the last ban when reaching the time, i still don't know how Rakete's bans got that timestamp... But, this is going to fix an issue related to another bans being removed.

**How to reproduce?**
- Start the server, do a ban ip in a random ip with the time of 1m (1 minute). Now ban other IPs, make a copy of the bans.txt for checking which bans got removed, then just restart the server after 1 minute. When checking the bans.txt and comparing with the copy, you will be able to notice the 1 minute ban still on there (what will cause more bans being vanished), and the last ban you did got removed. So, if you had more bans that got expired, more wrong bans was going to be removed.

**Some screenshots:**
* Console saying that removed 1 ban:
Screenshot_2022-06-16-08-51-40_1024x768

* New bans.txt (69.69.69.69 got banned for 1 minute):
Screenshot_2022-06-16-08-52-32_1024x768

* Old bans.txt:
Screenshot_2022-06-16-08-53-59_1024x768

* Permanent ban being removed:
Screenshot_2022-06-16-08-54-36_1024x768

### DIFF
--- a/piqueserver/networkdict.py
+++ b/piqueserver/networkdict.py
@@ -73,7 +73,10 @@ class NetworkDict:
         self.networks.pop(ip)
 
     def pop(self, *arg, **kw):
-        network, value = self.networks.pop(*arg, **kw)
+        if not arg or not kw:
+            network, value = self.networks.popitem()
+        else:
+            network, value = self.networks.pop(*arg, **kw)
         return get_cidr(network), value
 
     def iteritems(self):
--- a/piqueserver/networkdict.py
+++ b/piqueserver/networkdict.py
@@ -70,10 +70,10 @@ class NetworkDict:
 
     def __delitem__(self, key):
         ip = ip_network(str(key), strict=False)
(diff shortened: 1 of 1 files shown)


NotAFile (approved):

NotAFile:
Hm, this breaks the tests.

DryByte:
Ah... Lemme fix it, he is trying to do pop(), but in dicts need an argument. One sec.

NotAFile:
When you're fixing it, sure that `/undoban` still works, which is supposed to undo the last ban. That's what this test is there for.

DryByte:
Yes, i saw that now.

DryByte:
Now everything is supposed to be working fine.

NotAFile (approved):
I'm not a huge fan of overloading `pop()` with two meanings, but at least this works so I'll merge this now.

PR #737, opened 2022-12-03
Fix for votekick returning "Invalid Player" when trying to votekick players with # in name

DryByte:
This will fix a bug related to votekick script, where if a player joins with # in start of the nickname, was unable to get votekicked when server had ratio or blockinfo enabled (most servers does that). That happened because these scripts was giving the player's name instead of the player's id, so when get_player tries to find the player "###" it can't find since # is not an ID (this probably can be fixed in a future PR).

**Steps to reproduce:**
1. Enable the scripts: votekick, ratio and blockinfo (in this order).
2. Join in the server with some users, and one using # as his nickname
3. Try to votekick the # guy.

This probably is related to #709 , so that can be closed or merged.

### DIFF
--- a/piqueserver/scripts/blockinfo.py
+++ b/piqueserver/scripts/blockinfo.py
@@ -28,7 +28,8 @@ from piqueserver.commands import command, admin, get_player
 from piqueserver.config import config
 
 blockinfo_config = config.section("blockinfo")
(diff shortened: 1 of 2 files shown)


NotAFile (approved):
I think this is still a bit hacky, ideally the player ID should be passed as a number instead, but it does fix the issue for now.

PR #747, opened 2023-02-16
BetterSpades send_chat_* support

DryByte:
This adds support for custom messages pre built functions for showing text in screen.
This will add 4 new chat types in constants for the custom chats, and add a new parameter for send_chat function to support custom chat types when the custom type is greater than 2 (CHAT_SYSTEM).


pyspades/player.py
```diff
@@ -1217,11 +1217,21 @@ def continue_map_transfer(self) -> None:
...
+    def send_chat(self, value: str, global_message: bool = False, custom_type: int = CHAT_ALL) -> None:
         if self.deaf:
             return
         chat_message = loaders.ChatMessage()
-        if not global_message:
+        if custom_type > 2 and "client" in self.client_info:
+            if self.client_info["client"] == "BetterSpades":
+                chat_message.chat_type = custom_type
+
+            elif self.client_info["client"] == "OpenSpades":
```
NotAFile, line 1228: iirc the capability to send custom message types in OS predates the ability to query for client version.
Because of this, I would just unconditionally send the markup in front. It doesn't hurt to just send it either way.

pyspades/player.py
```diff
@@ -1217,11 +1217,21 @@ def continue_map_transfer(self) -> None:
...
+                value = OPENSPADES_CHATTYPES[custom_type]+value
```
NotAFile, line 1229: pep8
```suggestion
                value = OPENSPADES_CHATTYPES[custom_type] + value
```

pyspades/player.py
```diff
@@ -1217,11 +1217,21 @@ def continue_map_transfer(self) -> None:
...
         self.protocol.transport.write(data, self.address)
 
-    def send_chat(self, value: str, global_message: bool = False) -> None:
+    def send_chat(self, value: str, global_message: bool = False, custom_type: int = CHAT_ALL) -> None:
         if self.deaf:
             return
         chat_message = loaders.ChatMessage()
-        if not global_message:
+        if custom_type > 2 and "client" in self.client_info:
+            if self.client_info["client"] == "BetterSpades":
```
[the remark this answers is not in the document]
NotAFile: It seems that `player.proto_extensions` exists so I think all that should need to be done is to add a constant for  the chat extension and then query for that here...?

pyspades/player.py
DryByte: You mean without checking OS version, or for all clients?

NotAFile: I would send it for all clients that do not support the chat type extension?

pyspades/player.py
```diff
@@ -1217,11 +1217,20 @@ def continue_map_transfer(self) -> None:
...
-    def send_chat(self, value: str, global_message: bool = False) -> None:
+    def send_chat(self, value: str, global_message: bool = False, custom_type: int = CHAT_ALL) -> None:
         if self.deaf:
             return
         chat_message = loaders.ChatMessage()
-        if not global_message:
+        if custom_type > 2 and "client" in self.client_info:
+            if self.client_info["client"] == "OpenSpades":
+                value = OPENSPADES_CHATTYPES[custom_type] + value
+            elif EXTENSION_CHATTYPE in self.proto_extensions:
```
NotAFile, line 1227: I think the extension should be checked first because it should be preferred to the old OpenSpades way, but if we're going to send the OpenSpades markup to everyone then that's going to happen either way

NotAFile (approved):
LGTM in this stage already, up to you whether you want to change how the fallback is handled

pyspades/player.py
DryByte: Yes, i was checking for OS first because though of not sending the "prefix" for everyone, but since its just two characters, why not?

PR #748, opened 2023-02-16
Update in TDM

DryByte:
Add support to use config section and score percentage how it was discussed on https://github.com/piqueserver/piqueserver/discussions/739 you can use any capture limit for it, and i highly recommend to use 10 or 99, and set as true remove_intel option.

The captures are not added as real kills to the players, only the team score is used.

### DIFF
--- a/piqueserver/game_modes/tdm.py
+++ b/piqueserver/game_modes/tdm.py
@@ -1,12 +1,30 @@
 """
 Team Deathmatch game mode.
 
-Maintainer: Triplefox
+Options
+^^^^^^^
+.. code-block:: toml
+    [tdm]
+    # Maximum kills to win the game
+    kill_limit = 100
+
+    # How many points you will get by intel capture
+    intel_points = 10
+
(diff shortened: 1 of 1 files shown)


piqueserver/game_modes/tdm.py
```diff
@@ -65,30 +87,53 @@ def get_kill_count(self):
...
+            if SCORE_PERCENTAGE.get() and player:
+                team = player.team
+                caps_percent = math.floor(
+                    self.max_score*team.kills/KILL_LIMIT.get())
+
+                if caps_percent > team.score:
+                    self.do_captures(team, caps_percent)
+
+            if self.green_team.kills >= KILL_LIMIT.get():
+                self.broadcast_chat("Green Team Wins, %s - %s" %
```
NotAFile, line 130: It might actually be neat to switch this to use the actual team names in the future

NotAFile (approved):
lgtm

PR #749, opened 2023-02-20
Remove uses of protocol.send_chat and protocol.send_contained

DryByte:
This removes the uses of protocol.send_chat and protocol.send_contained, thats deprecated.
And autopep8 did some code changes, i changed some parts to fit with E501, but was too many and i was worried about breaking stuff, so i didnt fixed all E501. I think its better to fix E501 while we work in the files.


NotAFile:
Having two huge changes in one commit makes this impossible to review, so I can't really say anything about this...

DryByte:
Sorry, next time i will split the autopep8 part in another commit. But from my local tests everything was fine.

DryByte:
So, its better to delete last commit, and split it into 3 commits? One for send_chat, another for send_contained and a last one with autopep8?

NotAFile:
If that's easy for you to do it would be nice, it helps keep the history cleaner.

NotAFile (approved):
nice 🎉

PR #750, opened 2023-02-20
Add config options and player isolation in blue team

### LINKED ISSUE #211: Replace FFA with better FFA
The current free-for-all game mode stinks. It's basically just teamkilling enabled. There was a better one that showed all players as being on the opposite team.

DryByte:
This will add config options, so players can change the gamemode options inside their config. And add the option to isolate players in a different team, what actually happens is: server will set all players to green team, but when player tries to spawn server sends to the player a packet saying "hey you is on blue team", and for other players server sends "hey that dude is on green".

_(Also fixes an issue with team damage, that was not working right when having team damage on grief enabled)_

Fixes #211

### DIFF
--- a/piqueserver/game_modes/freeforall.py
+++ b/piqueserver/game_modes/freeforall.py
@@ -1,19 +1,30 @@
 """
 Free for All: shoot anyone
+
+Options
+^^^^^^^
+.. code-block:: toml
+    [freeforall]
+    # If ALWAYS_ENABLED is False, free for all can still be enabled in the map
+    # metadata by setting the key 'free_for_all' to True in the extensions
+    # dictionary
+    always_enabled = true
+
+    # If WATER_SPAWNS is True, then players can spawn in water
+    water_spawns = false
+..
 """
 # Free for all script written by Yourself
 
 from random import randint
 
+from pyspades.contained import CreatePlayer, ExistingPlayer
+from piqueserver.config import config
 from pyspades.constants import CTF_MODE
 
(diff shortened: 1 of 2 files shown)


piqueserver/game_modes/freeforall.py
```diff
@@ -1,19 +1,41 @@
...
+
+    # If WATER_SPAWNS is True, then players can spawn in water
+    water_spawns = false
+
+    # If ISOLATE_PLAYER is True, then player will be alone in
+    # the blue team, vs all greens auto disabling use_score_hack
+    isolate_player = true
+
+    # If USE_SCORE_HACK is True, then when you kill someone
+    # he will get changed to the other team, so you can
```
NotAFile, line 21: Avoid using "he" pronouns for the user, e.g.
```suggestion
    # that player will appear to be on the other team, so you can
```

piqueserver/game_modes/freeforall.py
```diff
@@ -52,16 +74,37 @@ def on_flag_spawn(self, x, y, z, flag, entity_id):
...
+                                                sender, team, save, rule)
+
     class FreeForAllConnection(connection):
         score_hack = False
 
+        def on_team_join(self, team):
+            if team.spectator or not ISOLATE_PLAYER.get():
+                return team
+
+            return self.protocol.green_team
```
NotAFile, line 99: I think there's a `.team_1` member variable to avoid the confusion with team names?

NotAFile (approved):
Looks good overall!

PR #755, opened 2023-02-28
Change python3.6 to python3.7

tydaytygx:
Change python3.6 to python3.7 since the server requires python3.7+ .
Tested on my server


NotAFile:
I feel like this should just be raised to the highest supported version for performance enhancements etc?

Issue #585, opened 2020-04-14
Review usage of reactor.seconds()


NotAFile:
There's many places in the code where Twisted's `reactor.seconds()` is used. However, the timestamps returned by that call are only guaranteed to be valid relative to other calls to `reactor.seconds()`.

The following should be done:
 - `world_time` used for game logic
 - `time.monotonic` used for durations, such as "280 seconds"
 - `time.time` used for absolute time, such as a "Tuesday 12pm"
 - `reactor.seconds` used for times relative to the event loop, such as `call_at(time)`

NotAFile:
Neat, it's been a few years, was wondering if we were still on the list :)

aoengin:
Hi, 
I tried to contact @NotAFile via e-mail and matrix but I could not get a response. Is there anyone who can explain why I should use time.time() in place of reactor.seconds()? I examined the doc of twisted and made several tests but I could not find any differences between these two methods therefore I'm not sure if a refactoring for that situation is necessary and where I should use each method.
Thanks in advance.

NotAFile:
Hi, @aoengin. Sorry, it's the weekend. I am busy with other stuff. I'll get back to you on this on Monday.

NotAFile:
@aoengin This issue was originally created in regards to another bug, where a change between reactors caused a lot of issues because one used `time.time()`, while another used `time.monotonic()`

`time.time()` represents some amount of seconds since 1970 UTC, while `time.monotonic()` is guaranteed to increase by one for every second that the computer has been running. This is important, for example, when your computer goes into sleep, your clock changes, or a leap second happens.

World time represents the amount of time the game world has been alive. This is important for example in the instance of lag, as less time might pass in the game world than in the real world.

These numbers should never be mixed. The initial issue showed that these numbers were being mixed in some cases. Hence the goal of this issue was to try to determine if there were any other of these cases.

tayloh:
We would like to create tests in relation to this issue, is that possible? Would you suggest any specific way to go about it?

NotAFile:
There's not really any way to test this, I'm afraid?

tayloh:
I figured that was the case, thanks for the quick response!

PR #757, opened 2023-03-08
Refactor usage of reactor.seconds() (#585)

tayloh:
Attempt at resolving #585 

We've made an attempt at refactoring the usage of reactor.seconds() in 19 files over 65 occurences (searched with VSCode), we think that's all of them.

Additionally, we made an attempt at creating tests that verify that the functionality in affected methods or functions (that we were able to test, which is not all of them), that depend on the timestamps, work as intended.


NotAFile:
Closing because it would be a lot of work to unpick the changes that we actually want and the contributors have already moved on

PR #762, opened 2023-03-24
Add a script to recover hp like COD

tydaytygx:
Tested on my server, simple function: add hp after hp is lower than 100 with delay


DryByte:
Hey, cool idea. But i dont think that new connection class is necessary, neither calling a list for player.values().

Suggestion: maybe its cool to add an option for recovering hp only if the player is crouching/sneaking and not in a fight in the last 30 seconds.

tydaytygx:
> But i dont think that new connection class is necessary



> Hey, cool idea. But i dont think that new connection class is necessary, neither calling a list for player.values().
> 
> Suggestion: maybe its cool to add an option for recovering hp only if the player is crouching/sneaking and not in a fight in the last 30 seconds.

Thanks for the suggestion! I will try to improve it.

NotAFile:
I think we'll also have to discuss whether there is enough demand for this to end up in the core repository. We've tried to cut down on the amount of scripts in the repository in the past and moved a bunch to -extra, but we don't really have any guidelines for accepting new ones yet

tydaytygx:
> I think we'll also have to discuss whether there is enough demand for this to end up in the core repository. We've tried to cut down on the amount of scripts in the repository in the past and moved a bunch to -extra, but we don't really have any guidelines for accepting new ones yet

Placing additional scripts to -extra repository is a good idea, and we should add the link in this main repository's README.md.

tydaytygx:
> Hey, cool idea. But i dont think that new connection class is necessary, neither calling a list for player.values().
> 
> Suggestion: maybe its cool to add an option for recovering hp only if the player is crouching/sneaking and not in a fight in the last 30 seconds.

I am wondering how to break the  hp recovery loop when player is under attack or getting damage. Which function should I use?
Maybe on_hit_recieved? But it seems not include grenade and falling damage.

DryByte:
You can add one for the 3 of them, but use the on_hit one instead of the on_hit_received. The received ones are used for direct interaction with packets without piqueserver validating.

Btw, i think this PR can be closed and reopened on piqueserver-extras. And for future questions you can use the discussion tab, the discord or the matrix chat.

NotAFile:
Closing as per the consensus to put this in -extras instead, as @DryByte said

I've opened #763 to hopefully make that more visible and easy to use soon.

PR #774, opened 2024-01-21
Make the server send it supported extensions to the client

DryByte:
Send to the client which extensions server supports, so client's can modify the experience.

### DIFF
--- a/piqueserver/server.py
+++ b/piqueserver/server.py
@@ -272,7 +272,7 @@ class FeatureProtocol(ServerProtocol):
         self.win_count = itertools.count(1)
         self.bans = NetworkDict()
 
-        self.proto_extensions = [(EXTENSION_CHATTYPE, 1)]
+        self.available_proto_extensions = [(EXTENSION_CHATTYPE, 1)]
 
         # attempt to load a saved bans list
         try:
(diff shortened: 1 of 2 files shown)


NotAFile:
I don't really see the point of this: We already have a good fallback for these message types. I don't see why admins would want to modify this in a config.

piqueserver/server.py
```diff
@@ -272,6 +272,8 @@ def __init__(self, interface: bytes, config_dict: Dict[str, Any]) -> None:
         self.win_count = itertools.count(1)
         self.bans = NetworkDict()
 
+        self.proto_extensions = [(EXTENSION_CHATTYPE, 1)]
```
NotAFile, line 275: I would make it clearer that this is the list of extensions the server *theoretically* supports. With it being the same variable name as the client where it is the things the client *actually* supports, it could get confusing.

NotAFile (changes requested):
Overall happy with this now, apart from theslightly confusing name

piqueserver/server.py
DryByte: Any name suggestion? I though about just changing it to "extensions", but i think it would lead to confusion with map extensions, or even with the extensions.py file.

I kept as proto_extensions, because it reminds me of protocol_extensions.

NotAFile: how about `available_proto_extensions`? My main worry is people accidentally checking if the *server* supports supports something instead of the client. If they have the same name that's very easy.

DryByte: That would work, i guess

NotAFile (approved):
lgtm!

PR #776, opened 2024-01-28
perf(mapgenerator): lower the compression level to 5

utf-4096:
This PR is a proposal to lower the compression level from 9 to 5.

Here are some tests I made from compressing Aloha's map repository (806 maps):

| Level | Average time | Average size |
| ----- | ------------ | ------------ |
| 9     | 325ms        | 510KiB       |
| 5     | 48ms         | 539KiB       |

Level 9 is only ~1.06x smaller than level 5 on average, but takes **6.7x as long** to compress.

Additionally, [SpadesX already uses level 5](https://github.com/SpadesX/SpadesX/blob/master/Source/Util/Compress.c#L66).

Much higher compression times for a very small decrease in size doesn't sound like it's worth it, so I think it's better to change it :)

### DIFF
--- a/pyspades/mapgenerator.py
+++ b/pyspades/mapgenerator.py
@@ -4,7 +4,7 @@ to the client on connect
 """
 import zlib
 
-COMPRESSION_LEVEL = 9
+COMPRESSION_LEVEL = 5
 
 
 class ProgressiveMapGenerator:


NotAFile:
Sounds sensible, especially with the general increase in bandwidth since 2012.

I recall playing around with this a while back and it not being a big difference in cpu usage or download times since it's on it's own thread anyway and we have some caching. But it doesn't hurt either.

NotAFile (approved):

PR #793, opened 2024-06-01
Truly hide the intel on pickup for onectf/babel

DavidCo113:
Copied over from https://github.com/aloha-pk/piqueserver/pull/46.

Sends the untaken intel off to {INFINITY, INFINITY, 128} instead of the top-left corner of the map when the other is taken, thereby making it impossible to see on the map.

Tested on OpenSpades v0.1.2/v0.1.3/v0.1.5, AoS v0.75, and BetterSpades 332a8eb54621eddf6e83253f9b0919d226902ae7. This method of hiding the intel is additionally used in the libspades EXPERIMENTAL server, and has not shown any issues there (that I know of) either.

### DIFF
--- a/piqueserver/game_modes/babel.py
+++ b/piqueserver/game_modes/babel.py
@@ -7,12 +7,13 @@ Release thread:
 http://www.buildandshoot.com/viewtopic.php?t=2586
 """
 
+import math
 from pyspades.constants import CTF_MODE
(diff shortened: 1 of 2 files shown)


NotAFile (approved):

utf-4096 (approved):
Tested working on OS 0.1.5, BS and Voxlap on my machine.

PR #795, opened 2024-07-15
Nix " Player was kicked: Rapid hack detected"

a7u8:
Removes the portion of the feature which kicks laggy players, and leaves the rest.


NotAFile:
What is the justification here? Do you have any data to back it up? What measures will prevent the cheats this prevents once this is removed?

DavidCo113:
The justification is that it always and often kicks laggy players and is minimally, if at all, effective against cheaters (my time on the FL Anarchy server should be proof of this). The laggy player kicking can be observed often in aloha.pk's tower of babel.

This change removes the kicking part and leaves behind the blocking part, that being the return statement. This prevents kicking laggy players, while also preventing cheaters from going completely overboard.

However, the timing in pyspades/constants.py is (and always was) too small, leading to 5x (in practice 2.5x) speed block place and 10x (in practice 5x) speed 1x3 dig, and 4x (in practice 2.35x) speed for 1x1 dig. This *should not* be significantly affected by this change in its current state. I am not certain if merely correcting the constants would truly fix the issue, though. Most likely, a different rate limiter (not to be confused with RateLimiter, I'm referring to the one in the if statements), such as one like my own used in libspades, would need to be implemented.

NotAFile:
Right, I see. The problem with that, and the reason we have this detection in place, is that there is a leniency between what a legitimate player can do and what the server is willing to accept, as tolerance for lag. If we don't do any rapid detection, cheating players would be able to perform those actions at the higher, maximum tolerance rate for sustained periods of time. We would then have to reduce those tolerances, which would reduce the kicking but then lead to many more cases of e.g. blocks not being placed due to lag.

I'd suggest that, since it does sound like it's a bit sensitive right now, we increase the threshold for rapid detection instead.

DavidCo113:
Probably about time I got back to this. I've made the timing more strict, but have not tested it. This is still not a perfect solution, though, since players with stable round-trip time should hypothetically be able to shave off at most 0.1 s per action.

pyspades/constants.py
```diff
@@ -72,9 +73,11 @@
 HIT_TOLERANCE = 5.0
 CLIP_TOLERANCE = 10
 
+SPADE_TOOL_3X = 4
```
NotAFile, line 76: I like the separation of rapid detection by tool. But I'm not a huge fan of having this value here. It's bound to be confusing because it's separate from the other tool values, only used by speedhack detection and invalid in the rest of the code.

I think there's two ways of possibly doing it better, depending on how much it interferes with the rest of the code: having a separate `enum.enum` for events that are rate limited, or getting rid of the dict and just having straightforward constants for the rate limits.

NotAFile:
@utf-4096 how does aloha handle this?

utf-4096:
Right now, default piqueserver behavior, which is annoying because all it does is kick players who lag and 1% of the time it kicks hackers (who just rejoin and use another hack).
Previously (PySnip era) we had a script that would prevent rapid hack by just discarding the packets that came in too fast (not by using deque IIRC, just checking last packet time) and also sent a notification to admins via IRC.

a7u8:
Any plans on ripping the kicking out?

Detank2002:
Plus one for removing the kicking here, all we have noticed at aloha is that the kick only really ends up with false positives, kicking lagging players repeatedly, it's a feature that could deal with the odd hacker but has a way too high error rate to justify it being a feature, nevermind default.

NotAFile (approved):
not the prettiest way to do this but it's gonna save people a lot of annoyance, so let's merge this finally

PR #797, opened 2024-08-04
fix(server): skip updating entities with an invalid position

utf-4096:
Related to #793

Trying to update entities that have invalid positions results in `OverflowError`s and in some cases hanging the server.

This PR skips updating entities that have a position that is invalid (which is when any axis is either infinity or nan).

### DIFF
--- a/pyspades/server.py
+++ b/pyspades/server.py
@@ -32,7 +32,7 @@ from pyspades.master import get_master_connection
 from pyspades.team import Team
 from pyspades.entities import Territory
 # importing tc_data is a quick hack since this file writes into it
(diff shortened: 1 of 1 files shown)


NotAFile:
I would still prefer to just remove entities that are not in play from the entities list, but this does fix some bugs for now

utf-4096:
Not sure I understand what you mean?

utf-4096:
Discussed privately: this isn't an ideal solution, it would be better to remove the intel we want to hide from the entities list entirely, and have the server decide the position to send. This PR isn't entirely useless however, in case we have some bugs that make some entities' position infinity/nan, this prevents the server from dying.

PR #800, opened 2024-11-11
feat(master): implement multiple master servers

utf-4096:
This PR adds the possibility of announcing the server to multiple master server lists, instead of the only previously-hardcoded buildandshoot list.

While buildandshoot has been immensely important to the survival of Ace of Spades, it has also been causing trouble to servers owners in various ways recently. Mainly:
- Lack of action for servers spoofing player count, spamming the server list or impersonating other servers
- No support for servers with >32 players
- Frequent downtimes (from my own measurements: at least 6 times since march this year)

We can't completely get rid of buildandshoot *right now* (or maybe ever), but we can start slowly transitioning away from it by also moving clients to an alternate serverlist (which would merge results from bns & other masterlists, so no one's left out).

Changing master servers has historically been a sensitive topic, so please leave any concerns you might have about this in the comments of this PR.

### DIFF
--- a/piqueserver/server.py
+++ b/piqueserver/server.py
@@ -64,7 +64,6 @@ from pyspades.vxl import VXLData
 
 log = Logger()
 
-
 def validate_team_name(name):
     if len(name) > 9:
         log.warn(
@@ -177,7 +176,11 @@ cmd_antispam_enable = config.option("enable_command_ratelimit", True)
 cmd_command_limit_size = config.option("command_ratelimit_amount", 4)
 cmd_command_limit_time = config.option(
     "command_ratelimit_period", "5s", cast=cast_duration)
-
+master_hosts = config.option("master_hosts", [
+    {'host': 'master.buildandshoot.com', 'port': 32886},
+    {'host': 'master1.aos.coffee', 'port': 32886},
+    {'host': 'master2.aos.coffee', 'port': 32886},
+])
 
 def ensure_dir_exists(filename: str) -> None:
     d = os.path.dirname(filename)
@@ -216,8 +219,8 @@ class FeatureProtocol(ServerProtocol):
     global_chat = True
     remote_console = None
     advance_call = None
-    master_reconnect_call = None
-    master = False
+    master: bool = False
+    master_hosts = []
     ip = None
     identifier = None
     command_antispam = False
@@ -338,6 +341,7 @@ class FeatureProtocol(ServerProtocol):
         self.command_antispam = cmd_antispam_enable.get()
         self.command_limit_size = cmd_command_limit_size.get()
         self.command_limit_time = cmd_command_limit_time.get()
+        self.master_hosts = master_hosts.get()
 
         # voting configuration
         self.default_ban_time = default_ban_duration.get()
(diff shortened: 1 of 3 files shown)


piqueserver/config/config.toml
```diff
@@ -38,6 +38,13 @@ tips_frequency = "5min"
 # set this if the server should register on the master server list
 master = false
 
+# configure multiple master servers here
+# NOTE: you should leave at least buildandshoot for your server
+# to be viewable by everyone
+master_hosts = [
```
NotAFile, line 44: I would comment this out my default so that we can can change the defaults in future updates without having to make everyone change their configs.

The further thing worth debating is whether we want to feature this so prominently in the settings as few users will want to change this and people changing it not knowing the ramifications may increase fragmentation.

DryByte:
I think this is a good idea since we can't guarantee buildandshoot will be alive for a long time or even get better support.
So, would be cool if we get a way to support "extensions" for these new masterlists, i.e.:
- 32+ player support
- Server logo
- Longer server description

This should be documented and discussed on https://github.com/piqueserver/aosprotocol

Also, for now i think would be good if we kept buildandshoot as a "required" masterlist to connect, so we don't get headaches of people reporting errors due the deletion of the masterlist connection in config file, and not starts a full transaction of the masterlist, with servers going to obviously better masterlists and leaving bns empty.

The issue with leaving bns empty is that bns still is the main gate for new players, if you download openspades/betterspades it comes automatically set to bns, this probably will be easily fixed using forks (or even a PR), as an example we have IV of Spades which got really popular in the last days. So, leaving bns empty will not help to make the game grow or for old players come back.

NotAFile (commented):
I have been very skeptical of this idea in the past as it seemed like opening pandora box, but with the current state of the master server this seems like it has to happen now and will cause the least disruption for players long term.

In my opinion we should be shipping this sooner rather than later, with a trusted alternative master server list by default alongside the current one.

piqueserver/config/config.toml
utf-4096: Fair point, it might be better to remove it from the example config entirely. Would definitely avoid users asking for help in not understanding why their server isn't showing up on the masterlist because they changed it by accident.

utf-4096:
@DryByte

> I think this is a good idea since we can't guarantee buildandshoot will be alive for a long time or even get better support. So, would be cool if we get a way to support "extensions" for these new masterlists, i.e.:
> * 32+ player support
> * Server logo
> * Longer server description
>
> This should be documented and discussed on [piqueserver/aosprotocol](https://github.com/piqueserver/aosprotocol)

Good ideas but out of the scope of this PR.

> Also, for now i think would be good if we kept buildandshoot as a "required" masterlist to connect, so we don't get headaches of people reporting errors due the deletion of the masterlist connection in config file, and not starts a full transaction of the masterlist, with servers going to obviously better masterlists and leaving bns empty.
> 
> The issue with leaving bns empty is that bns still is the main gate for new players, if you download openspades/betterspades it comes automatically set to bns, this probably will be easily fixed using forks (or even a PR), as an example we have IV of Spades which got really popular in the last days. So, leaving bns empty will not help to make the game grow or for old players come back.

Not sure I completely understand what you're saying, but I think its still a good idea to give users who understand what they're doing the option to remove BnS and have their own private masterlist (one use case I can think of is aloha test servers). It's in the best interest of server owners to stay leave BnS enabled anyway (why wouldn't you want more players?)

I've removed the option from the example `config.toml`, this should avoid most users from mistakenly changing it.

NotAFile (approved):

NotAFile (approved):

PR #802, opened 2024-12-31
fix(master): send player count on connect when `count > 0`

utf-4096:
Not sending the player count on startup is fine because it is always zero, but if there are players on the server and you reconnect to a master server, the player count won't be updated until someone joins/leaves.

Reported by coherent.sheaf.

### DIFF
--- a/pyspades/master.py
+++ b/pyspades/master.py
@@ -168,6 +168,9 @@ class MasterPool:
         self.descriptors = []
 
     def on_master_connect(self, client: MasterConnection):
+        if (count := self.protocol.get_player_count()) > 0:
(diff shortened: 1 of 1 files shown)


NotAFile:
Is there any disadvantage to just sending it unconditionally? But lgtm as is too.

NotAFile (approved):

utf-4096:
Not really any disadvantage, just that sending a CountUpdate is unnecessary since the player count is already 0 when you connect for the first time.

NotAFile:
Yeah, it's just that having less code paths makes things easier to test and reason about. But your call.

PR #803, opened 2024-12-31
fix(scripts/aimbot2): divide shotgun hits by 8

utf-4096:
The shotgun does 8 hits per fire instead of rifle/SMG which only do 1.
This fixes aimbot2 misreporting 100% accuracy as 800%.


NotAFile (approved):
code itself looks good to me, I don't know how robust this is but it seems preferrable to spamming the logs with false positives

PR #804, opened 2025-01-06
fix(player): replace `\r` and `\n` in chat messages with a space

utf-4096:
This avoids some minor annoyances with IRC and OpenSpades users.

### DIFF
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -37,6 +37,8 @@ log = Logger()
 
 tc_data = loaders.TCState()
 
+# special characters to replace in chat messages
+MSG_SPECIAL_CHARACTER_MAP = str.maketrans({'\r': ' ', '\n': ' '})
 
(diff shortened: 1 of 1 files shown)


pyspades/player.py
```diff
@@ -635,7 +635,11 @@ def on_chat_message_recieved(self, contained: loaders.ChatMessage) -> None:
         if not self.name:
             return
 
-        value = contained.value
+        value: str = contained.value
+
+        charmap = str.maketrans({'\r': ' ', '\n': ' '})
```
NotAFile, line 640: This is a bit wasteful, you could construct it once and then reuse it.

NotAFile (approved):

pyspades/player.py
NotAFile: I'd just put it in the same file probably. No need to export it more widely.

PR #811, opened 2025-03-06
fix(master): catch dns lookup failures inside `add_client()`

utf-4096:
We were already catching DNS lookup failures, but only on first connection, not on subsequent reconnects.

This can happen if the server starts with a working internet connection, but then loses connection for over 60 seconds (the reconnect interval). Since the reconnect call is not using try-except, we get an uncaught error.

This PR moves the try-except inside `add_client` directly so we catch it on first connect and on any reconnect.

---

Stack trace of the error that brought up this bug (noticed on aloha servers):
```py
2025-03-03T18:14:31+0000 [twisted.internet.base#critical] while handling timed call
          File "/app/venv/lib/python3.12/site-packages/twisted/internet/base.py", line 1105, in runUntilCurrent
            call.func(*call.args, **call.kw)
          File "/app/venv/lib/python3.12/site-packages/pyspades/master.py", line 194, in <lambda>
            reactor.callLater(60, lambda: self.add_client(client.descriptor))
          File "/app/venv/lib/python3.12/site-packages/pyspades/master.py", line 131, in add_client
            connection = self.protocol.connect(
          File "/app/venv/lib/python3.12/site-packages/pyspades/protocol.py", line 99, in connect
            peer = self.host.connect(enet.Address(host, port), channel_count,
          File "enet.pyx", line 265, in enet.Address.__init__
            
          File "enet.pyx", line 300, in enet.Address.host.__set__
            
        builtins.OSError: Resolution failure!
```

### DIFF
--- a/pyspades/master.py
+++ b/pyspades/master.py
@@ -128,15 +128,19 @@ class MasterPool:
         self.descriptors += [descriptor]
 
     def add_client(self, desc: MasterHostDescriptor):
-        connection = self.protocol.connect(
-            MasterConnection, desc.host, desc.port, MASTER_VERSION)
+        try:
+            connection = self.protocol.connect(
+                MasterConnection, desc.host, desc.port, MASTER_VERSION)
+        except OSError as error:
+            self.on_master_disconnect(client=None, descriptor=desc, reason=repr(error))
+            return
 
         connection.server_protocol = self.protocol
         connection.descriptor = desc
         connection.on_master_connect = \
             lambda: self.on_master_connect(connection)
         connection.on_master_disconnect = \
-            lambda: self.on_master_disconnect(connection)
+            lambda: self.on_master_disconnect(connection, desc, reason='disconnected')
 
         self.clients += [connection]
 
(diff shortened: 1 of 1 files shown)


pyspades/master.py
```diff
@@ -128,15 +128,20 @@ def add_descriptor(self, host: str, port: int):
         self.descriptors += [descriptor]
 
     def add_client(self, desc: MasterHostDescriptor):
-        connection = self.protocol.connect(
-            MasterConnection, desc.host, desc.port, MASTER_VERSION)
+        try:
+            connection = self.protocol.connect(
+                MasterConnection, desc.host, desc.port, MASTER_VERSION)
+        except OSError:
```
NotAFile, line 134: I'm not a big fan of just silently swallowing something as broad as OSError... I like the idea of unifying the reconnect and initial connect logic, but losing the ability to debug things at all is not great.

NotAFile (changes requested):

pyspades/master.py
NotAFile: Yeah I would do it at higher log level though because it's something end users are probably going to want to see. Ideally combined with the other log message like before?

NotAFile (approved):

PR #813, opened 2025-04-16
Included support in the broadcast_chat

MSGanesh23:
Edited player.py file to include the support in broadcast_chat

### DIFF
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -632,42 +632,46 @@ class ServerConnection(BaseConnection):
         self.protocol.broadcast_contained(contained, save=True)
         self.protocol.update_entities()
 
-    @register_packet_handler(loaders.ChatMessage)
-    def on_chat_message_recieved(self, contained: loaders.ChatMessage) -> None:
-        if not self.name:
+@register_packet_handler(loaders.ChatMessage)
+def on_chat_message_recieved(self, contained: loaders.ChatMessage) -> None:
+    if not self.name:
+        return
+
+    value: str = contained.value
+    value = value.translate(MSG_SPECIAL_CHARACTER_MAP)
+
+    if len(value) > 108:
+        log.info("TOO LONG MESSAGE ({chars} chars) FROM {name} (#{id})",
+                 chars=len(value),
+                 name=self.name,
+                 id=self.player_id)
+
+    value = value[:108]
+    if value.startswith('/'):
+        self.on_command(*parse_command(value[1:]))
+    else:
+        global_message = contained.chat_type == CHAT_ALL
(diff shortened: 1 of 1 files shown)


DryByte:
Hey, first of all, thanks for taking interest in contributing to piqueserver!

Taking a look into the changes, it will not fix the issue and has a lot of code that doesn't make sense and feels like you didn't tested the code before sending it.

The function "on_chat_message_recieved" its an event for player's sent messages, in other words, if we use that function other players will be able to send custom message types to other user's (which we don't want to happen, since that can become really annoy and "spammable")

Other issue with the code is that it can't run, if you try to test the code (or actually just join the server), it will break because some issues that i will mention later on, but the main issue is the indentation, seems like you copy pasted the code there and forget to "tab" it, so when python's interpreter tries to do the _MaGiC_, it breaks (not letting you join the server and throwing a random error in the console).

Now getting in the actual issue and why that will not really fix it, in the following line you added:
```py
contained.chat_type = contained.custom_type if hasattr(contained, 'custom_type') else CHAT_ALL
```
Not makes any sense, because "custom_type" isn't a part of the packet, the packet uses chat_type for the custom types handling: https://www.piqueserver.org/aosprotocol/extensions.html#message-types


So, if you really want to help and contribute, i highly recommend you to actually write the code and test it before hand. If you have any questions, related to the protocol, or even the piqueserver structure, you can join in our discord or in the matrix room.

Just to append some informations here if you want to try again:
- The function we are talking is the "broadcast_chat", that's located in pyspades/server.py, that's a function that exists for piqueserver internals and scripts usage. If you take a look in that file you can see some functions called: "broadcast_chat_warning/notice/error/status", those functions can only send openspades's custom type handling, that uses a chat prefix, and not the packetless implementation, in those cases you can do what i did in the send_chat, using OPENSPADES_CHATTYPES to get the same result using the packetless implementation chat_type ID's.
- To understand better check the protocol documentation pages:
https://www.piqueserver.org/aosprotocol/protocol075.html#chat-message
https://www.piqueserver.org/aosprotocol/extensions.html#message-types
https://github.com/piqueserver/piqueserver/blob/master/pyspades/contained.pyx#L658

So, my final thoughts on this is, it can't actually run, and if you want to use AI, please test it before hand.

Piqueserver discord: https://discord.gg/w6Te7xC
Piqueserver matrix room: #piqueserver:matrix.org

NotAFile (changes requested):
What drybyte said. This code is, to me, obviously non-functional and parts of it are likely AI generated, which makes any other changes suspect and tainted too.

MSGanesh23:
Sorry for using AI 

I will try to write the code and test it before requesting PR

PR #814, opened 2025-04-16
Improve TOW gamemode

ashy1227:
- Add support for setting custom capture points in the map.txt
- Add support for odd numbers of capture points (The center capture point will be neutral, and either team will have to capture it before they're able to capture each other's territories)
- Add option to have a fixed spawnpoint instead of spawning at the second-closest one to the enemy
- Ability to use custom spawning logic in the map.txt by adding `'tow_use_custom_spawn_function': True` to the extensions


NotAFile:
Just from a brief look, cool stuff, thanks!

utf-4096 (approved):
LGTM

PR #819, opened 2025-04-18
Modernize build instructions

NotAFile:
- remove obsolete manylinux wheels
- replace deprecated uses of setup.py
- correct whatever other misinfo I found


Dockerfile
utf-4096, line 1: I think you can omit this change, since the Dockerfile has several other issues anyway. The https://github.com/piqueserver/piqueserver/tree/update-dockerfile branch fixes this so after I clean it up I'll make a PR.

doc/installation.rst
```diff
@@ -33,10 +33,8 @@ Installing from source
     # create a new python3 venv
     virtualenv -p python3 venv
```
utf-4096, line 34: Is this still the recommended way to create a venv? I've been using `python3.X -m venv venv` for quite a while, but maybe it's just a me thing.

doc/installation.rst
```diff
@@ -65,5 +63,5 @@ Tricky bit for windows is to get Cython working.
 
     cd piqueserver
     pip3 install -r requirements.txt
```
utf-4096, line 65: Not needed anymore.

doc/release-guide.rst
```diff
@@ -15,116 +15,51 @@ Ensure we’re ready for release
...
-    git commit -m "release version X"
-    git tag -a "v0.1.2"
-    git push
-    git push --tags
-
--  merge master across to the release branch if required:
-
-::
+    cd dist/
+    gh release dowload <git tag>
```
utf-4096, line 57:
```suggestion
    gh release download <git tag>
```

utf-4096 (changes requested):

doc/installation.rst
NotAFile: It depends on personal preference I think. But if people have preferences, they'll know how to create it differently.

```suggestion
    python3 -m venv venv
```

doc/installation.rst
NotAFile: Yeah this is actually still mentioned in a bunch of places, I'll clean that up

doc/installation.rst
utf-4096: On my current install the `virtualenv` command doesn't exist, but `python3 -m venv` does. So yeah `python3 -m venv venv` is the way to go.

NotAFile: Yeah you have to install it seperately. It's generally newer and more feature-complete than the stdlib one.

utf-4096 (approved):

PR #830, opened 2025-05-29
feat: improve handling of special characters in messages/names

utf-4096:
Changes made:
- Made messages get dropped when containing special characters instead of replacing with a space
- Replaced `.replace()` with `.translate()` for handling special characters in player names
- Added `#`, `\r` and `\n` as special characters in player names
- Made empty names get replaced with "Deuce" instead of leaving them blank

Replaces #709

### DIFF
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -37,8 +37,8 @@ log = Logger()
 
 tc_data = loaders.TCState()
 
-# special characters to replace in chat messages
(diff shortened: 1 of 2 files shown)


NotAFile (approved):
not sure why they need to be dropped instead of replaced as before but code looks good

utf-4096:
> not sure why they need to be dropped

Because I changed my mind and I think there's no point being tolerant with something you can't write on any client

PR #832, opened 2025-06-06
feat(server): print local aos identifier on start

utf-4096:
Adds the following log to the console:
```
[piqueserver.server#info] Local aos identifier: aos://16777343:32887
```

Useful for people who don't know the "magic" aos identifier for localhost.


piqueserver/server.py
```diff
@@ -423,7 +422,14 @@ def _post_init(self):
         self.master = register_master_option.get()
         self.set_master()
 
-    async def get_external_ip(self, ip_getter: str) -> Iterator[Deferred]:
```
utf-4096, line 426: I removed the weird `Iterator[Deferred]` return type because it was wrong.

PR #843, opened 2026-04-25
Revert "feat(player): drop messages with special chars instead of replacing"

a7u8:
This reverts commit 28038b9ee2586399939c64e64f5faf7c2cec6dac.

It's very easy to copy line feeds into the clipboard, ignoring a player's message because of that does not make that player very happy.

### DIFF
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -37,8 +37,8 @@ log = Logger()
 
 tc_data = loaders.TCState()
 
-# messages that contain these characters will be dropped
-BANNED_MESSAGE_CHARACTERS = ['\r', '\n']
+# special characters to replace in chat messages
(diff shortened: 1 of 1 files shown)


utf-4096:
Ugh, I forgot that OpenSpades lets you paste newlines. This is definitely something that should be fixed in clients and not worked around by us. But its not like we don't already have a bunch of fixes for broken clients in our server code, so...

Fran6nd:
@utf-4096 I joined zerospades/s maintainer as a co maintainer. Should we disable this ability to send mutiline chat message? it is not directly possible but I just checked, it is possible to past multiline text in chat input. I can fix this. Please confirm.

utf-4096:
Yes please.

PR #841, opened 2026-04-17
Provide a base bot API to ease bot integration into servers.

Fran6nd:
All is in the title :) 
Les me know if anything is wrong! Will be happy to make changes.

### New: Bot API (`piqueserver/bot.py`)

- **`FakePeer`** — drop-in `enet.Peer` stub; absorbs all network I/O
  so bots have no network presence
- **`BotConnection`** — headless `ServerConnection` that registers in
  `protocol.connections` and `protocol.players` like a real player;
  silences `send_contained`, `loader_received`, and map transfer
- **`Bot`** — high-level AI controller; subclass and override `think(dt)`
  for per-tick logic. Public API:
  - Positioning: `move_to()`, `look_toward()`, `look_horizontal_toward()`,
    `set_walk()`
  - Combat: `shoot_at()`, `throw_grenade()`
  - Building: `build_block()`, `destroy_block()`
  - Communication: `chat()`
  - Queries: `can_see()`, `distance_to()`, `get_enemies()`, `closest()`
- **`BotManagerMixin`** — protocol mixin for automatic per-tick bot
  management; mix in before the protocol class:
  ```python
  class MyProtocol(BotManagerMixin, protocol): ...

PR #844, opened 2026-04-30
implement AOS protocol kick reason extension (0xC2)

Fran6nd:
Implements the kick reason extension from the AOS 0.75 protocol
(piqueserver/aosprotocol#15). When a client advertises support for 0xC2,
the server now sends the kick/ban reason as a ChatMessage with
player_id=255 and chat_type=2 right before the ENet disconnect.

Compatible clients (BetterSpades, zerospades...) pick it up and display it
on the disconnect screen instead of the canned localized "You were kicked
from this server" text. Clients without the extension are unaffected —
the helper short-circuits and they fall back to existing behaviour.

Wired into kick() and the two on-connect ban paths in
piqueserver/player.py. The wire format helper send_kick_reason() lives on
ServerConnection in pyspades/player.py so any subclass can call it.

Tested against [zerospades](https://github.com/siecvi/zerospades): `/kick foo some reason` now shows `some reason`
on the disconnect screen.

My second PR after [this one](https://github.com/piqueserver/piqueserver/pull/841) (not yet merged) 😉

As new protocol extension are in discussion on the aloha discord, I thought it would be better to already implement existing ones first !

### DIFF
--- a/piqueserver/player.py
+++ b/piqueserver/player.py
@@ -66,7 +66,7 @@ class FeatureConnection(ServerConnection):
                 log.info('banned user {name} ({client_ip}) attempted to join',
                          name=name,
                          client_ip=client_ip)
-                self.disconnect(ERROR_BANNED)
+                self.disconnect(ERROR_BANNED, reason)
                 return
 
         manager = self.protocol.ban_manager
@@ -80,7 +80,7 @@ class FeatureConnection(ServerConnection):
                     client_ip=client_ip,
                     reason=reason
                 )
-                self.disconnect(ERROR_BANNED)
+                self.disconnect(ERROR_BANNED, reason)
                 return
 
         ServerConnection.on_connect(self)
@@ -333,7 +333,7 @@ class FeatureConnection(ServerConnection):
             log.info('{message}', message=message)
         # FIXME: Client should handle disconnect events the same way in both
         # main and initial loading network loops
-        self.disconnect(ERROR_KICKED)
+        self.disconnect(ERROR_KICKED, reason)
 
     def ban(self, reason=None, duration=None):
         reason = ': ' + reason if reason is not None else ''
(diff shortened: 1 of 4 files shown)


NotAFile (approved):

utf-4096:
I feel like an optional `reason` parameter for `disconnect()` might be better suited for the task rather than adding an extra call every time we want to disconnect.

Fran6nd:
Fair enough!

Fran6nd:
This way?

pyspades/constants.py
```diff
@@ -31,10 +31,20 @@
...
+ERROR_FULL = 4
+ERROR_SHUTDOWN = 5
+ERROR_KICKED = 10
+ERROR_INVALID_NAME = 20
+# AOS protocol extensions. piqueserver advertises EXTENSION_CHATTYPE and
+# EXTENSION_KICKREASON via FeatureProtocol.available_proto_extensions;
+# EXTENSION_PLAYERLIMIT is reserved but not implemented server-side.
+EXTENSION_PLAYERLIMIT = 192
+EXTENSION_CHATTYPE = 193
+EXTENSION_KICKREASON = 194
```
utf-4096: Please avoid style changes in this PR. If you want to reformat files, that can be a separate PR.

utf-4096 (changes requested):
Looks good, just one minor style issue.
Please also squash your commits afterwards.

NotAFile (approved):
yeah this is much nicer

pyspades/constants.py
NotAFile: I personally like this style change. It gets reall confusing when you have multiple range() things for one group of consfants. And it's a separate commit so it's fine to me. I don't think it's worth splitting PRs for that.

utf-4096: I don't think the change itself is bad. But it would be better if all files using `range()` were reformatted at once, rather than one small portion of this file in a feature PR.

utf-4096: I'll do it on master directly.

NotAFile: If you just don't squash the commits it'll be the same thing and less work, imo. But I don't care much either way.

utf-4096:
Thanks!

PR #847, opened 2026-05-03
Proposal and PR, allow built in mandatory extension check.

Fran6nd:
Adds required_proto_extensions on the protocol. Servers can list (extension_id, minimum_version) pairs that every client must advertise during the handshake — anyone missing one (or running too old a version) gets a chat error and is kicked before they can join.

⚠️ Clients not supporting a required extension and not supporting kick reason will be kicked without knowing why. This is the ugly part. A good thing would be to have the server list to know the server is requiring extension.

The check runs in two spots: as soon as the client's ProtocolExtensionInfo arrives (fast path for extension-aware clients), and again at ExistingPlayer time as a fallback for clients that never sent one (vanilla 0.75, older OpenSpades). Local connections are exempt.

Also lifts the long-standing skip of ProtocolExtensionInfo for OpenSpades ≤ 0.1.3 — but only when the server actually requires extensions, so those clients get the chance to advertise support before being judged.

Exposed via config.toml as required_proto_extensions = [[193, 1]]. Defaults to empty, behaviour is unchanged for everyone who doesn't opt in. The example in the shipped config is commented out.

Why
Right now extensions are just opportunistic — used for chat type colouring and not much else, so a "mandatory" knob isn't urgently needed.  Because as the protocol grows, extensions could start carrying actual gameplay-affecting features, servers will want a way to say "if your client doesn't speak X, you can't play here" without writing a custom script for it.

Note on testing
I'm on macOS arm64 and don't have a working build of any client old enough to be missing the extension protocol — vanilla 0.75 in particular. The "no ProtocolExtensionInfo ever arrives" path is therefore only verified by reading the code, not by actually connecting such a client. If anyone can spin up a vanilla 0.75 (or pre-extension OpenSpades) against this branch and confirm the kick fires at ExistingPlayer time as expected, I'd really appreciate it.

I do not know what you will think of it, let me know 😉 there are pros and cons.
For now people think more about using protocol extension vs new protocol version. so sooner or later there will be stuff affecting the gameplay coming out?

BTW, I can of course provide CI testing of it if you agree with the idea.

### DIFF
--- a/piqueserver/config/config.toml
+++ b/piqueserver/config/config.toml
@@ -85,6 +85,20 @@ scripts = [
 # number of /login attempts allowed before players are auto-kicked (default: 1)
 login_retries = 3
 
+# per-extension state for the protocol extensions piqueserver knows about.
+# any known extension you don't list here is treated as "disabled".
+# valid states:
+#   "enabled"  - server advertises it; clients may use it.
+#   "disabled" - server doesn't advertise it.
+#   "required" - server advertises it and kicks clients that don't
+#                support it back. clients with the "kick reason"
+#                extension see a descriptive message; others get the
+#                generic wrong-version dialog.
+proto_extensions = [
+  ["chat type", "enabled"],
+  ["kick reason", "enabled"],
+]
+
 # MAP
 # maps listed in the rotation
 # these names are the name of the map vxl/txt files without the file extensions
--- a/piqueserver/config/config.toml
+++ b/piqueserver/config/config.toml
@@ -87,18 +87,6 @@ login_retries = 3
 
 # per-extension state for the protocol extensions piqueserver knows about.
 # any known extension you don't list here is treated as "disabled".
-# valid states:
-#   "enabled"  - server advertises it; clients may use it.
-#   "disabled" - server doesn't advertise it.
-#   "required" - server advertises it and kicks clients that don't
-#                support it back. clients with the "kick reason"
-#                extension see a descriptive message; others get the
-#                generic wrong-version dialog.
-proto_extensions = [
-  ["chat type", "enabled"],
-  ["kick reason", "enabled"],
-]
-
 # MAP
 # maps listed in the rotation
 # these names are the name of the map vxl/txt files without the file extensions
--- a/piqueserver/config/config.toml
+++ b/piqueserver/config/config.toml
@@ -85,8 +85,6 @@ scripts = [
 # number of /login attempts allowed before players are auto-kicked (default: 1)
 login_retries = 3
 
-# per-extension state for the protocol extensions piqueserver knows about.
-# any known extension you don't list here is treated as "disabled".
 # MAP
 # maps listed in the rotation
 # these names are the name of the map vxl/txt files without the file extensions
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -103,6 +103,7 @@ class ServerConnection(BaseConnection):
         self.rapids = RateLimiter(RAPID_WINDOW_ENTRIES, MAX_RAPID_SPEED)
         self.client_info = {}
         self.proto_extensions = {}  # type: Dict[int, int]
+        self._enabled_ext_warned = False
         self.line_build_start_pos = None
 
     def on_connect(self) -> None:
@@ -144,12 +145,21 @@ class ServerConnection(BaseConnection):
         log.debug("received extinfo {extinfo} from {player}",
                   extinfo=self.proto_extensions,
                   player=self)
-        self._enforce_mandatory_extensions()
+        if self._enforce_mandatory_extensions():
+            return
+        self._warn_missing_enabled_extensions()
 
     def _missing_mandatory_extensions(self):
         return [
-            (ext_id, min_ver)
-            for ext_id, min_ver in self.protocol.mandatory_proto_extensions
+            (ext_id, min_ver, reason, name)
+            for ext_id, min_ver, reason, name in self.protocol.extensions.mandatory()
+            if self.proto_extensions.get(ext_id, -1) < min_ver
+        ]
+
+    def _missing_enabled_extensions(self):
+        return [
+            (ext_id, min_ver, reason, name)
+            for ext_id, min_ver, reason, name in self.protocol.extensions.enabled_only()
             if self.proto_extensions.get(ext_id, -1) < min_ver
         ]
 
@@ -164,14 +174,15 @@ class ServerConnection(BaseConnection):
         if not missing:
             return False
         log.info("{player} kicked: missing mandatory protocol extensions {missing}",
-                 player=self, missing=missing)
+                 player=self,
+                 missing=[(name, min_ver, reason)
+                          for _, min_ver, reason, name in missing])
         if EXTENSION_KICKREASON in self.proto_extensions:
             self.disconnect(
                 ERROR_KICKED,
                 reason="Missing mandatory client extension: " + ", ".join(
-                    "{} v{}".format(EXTENSION_NAMES.get(ext_id, "id %d" % ext_id),
-                                    min_ver)
-                    for ext_id, min_ver in missing
+                    "{} v{} ({})".format(name, min_ver, reason)
+                    for _, min_ver, reason, name in missing
                 ),
             )
         else:
@@ -181,6 +192,35 @@ class ServerConnection(BaseConnection):
             # this server don't agree on the protocol".
             self.disconnect(ERROR_WRONG_VERSION)
         return True
+
+    def _warn_missing_enabled_extensions(self) -> None:
+        """Tell the player about each ENABLED-but-missing extension once.
+
+        Only fires for ENABLED extensions, not MANDATORY ones (those caused
+        a kick instead). Sent once per connection; cheap if there's nothing
+        to warn about.
+        """
+        if self.local or self.disconnected or self._enabled_ext_warned:
+            return
+        missing = self._missing_enabled_extensions()
+        if not missing:
+            return
+        self._enabled_ext_warned = True
+        for ext_id, min_ver, reason, name in missing:
+            client_ver = self.proto_extensions.get(ext_id)
+            version_note = ""
+            if client_ver is not None:
+                version_note = " (your client: v{}, needed: v{})".format(
+                    client_ver, min_ver)
+            log.info("{player}: missing enabled extension {name!r}{ver} ({reason})",
+                     player=self, name=name, ver=version_note, reason=reason)
+            self.send_chat_warning(
+                'Heads up: this server uses the "{name}" protocol '
+                'extension ({reason}). Your client lacks it{ver}, so '
+                "related features won't work for you. Please update "
+                "your client, or open a ticket with its authors to add "
+                "support.".format(name=name, reason=reason,
+                                  ver=version_note))
 
     @register_packet_handler(loaders.ExistingPlayer)
     @register_packet_handler(loaders.ShortPlayerData)
@@ -201,6 +241,7 @@ class ServerConnection(BaseConnection):
         # (e.g. vanilla 0.75) — by now we've waited long enough for one.
         if self._enforce_mandatory_extensions():
             return
+        self._warn_missing_enabled_extensions()
 
         old_team = self.team
         team = self.protocol.teams[contained.team]
@@ -772,12 +813,12 @@ class ServerConnection(BaseConnection):
         # whether to kick them.
         skip_old_openspades = (contained.client == 'o'
                                and contained.version <= (0, 1, 3)
-                               and not self.protocol.mandatory_proto_extensions)
+                               and not self.protocol.extensions.mandatory())
         if skip_old_openspades:
             log.debug("not sending version request to OpenSpades <= 0.1.3")
         else:
             ext_info = loaders.ProtocolExtensionInfo()
-            ext_info.extensions = self.protocol.available_proto_extensions
+            ext_info.extensions = self.protocol.extensions.advertised()
             self.send_contained(ext_info)
         self.on_client_info()
 
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -218,7 +218,11 @@ class ServerConnection(BaseConnection):
                     client_ver, min_ver)
             log.info("{player}: missing enabled extension {name!r}{ver} ({reason})",
                      player=self, name=name, ver=version_note, reason=reason)
-            self.send_chat_warning(
+            # plain CHAT_SYSTEM rather than CHAT_WARNING: yellow toasts are
+            # disruptive at first spawn, especially when several extensions
+            # are missing -- chat-line delivery is less annoying and stays
+            # readable in scrollback.
+            self.send_chat(
                 'Heads up: this server uses the "{name}" protocol '
                 'extension ({reason}). Your client lacks it{ver}, so '
                 "related features won't work for you. Please update "
--- a/pyspades/player.py
+++ b/pyspades/player.py
@@ -164,13 +164,13 @@ class ServerConnection(BaseConnection):
     def _enforce_mandatory_extensions(self) -> bool:
         """Kick the player if any mandatory extension is missing.
 
-        Returns True if a kick was issued. Uses ERROR_WRONG_VERSION rather
-        than ERROR_KICKED with a custom reason: the client hasn't reached
-        the state where it processes ChatMessage packets at this point in
-        the handshake, so any custom kick-reason text we sent would be
-        dropped on the floor. "Incompatible client protocol version" is
-        the next-best built-in dialog and accurately describes the
-        situation.
+        Returns True if a kick was issued. Falls back to ERROR_WRONG_VERSION
+        because vanilla and v1 kick-reason clients do not process the
+        kick-reason ChatMessage during the handshake. For clients that
+        negotiated kick-reason v2 (see AOS extensions spec, packet 0xC2),
+        we additionally send a descriptive CHAT_SYSTEM reason before the
+        disconnect so the failure surfaces in the client UI instead of as
+        a generic "incompatible version" dialog.
         """
         if self.local or self.disconnected:
             return False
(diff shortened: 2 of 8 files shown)


NotAFile:
When I was mulling over this idea I decided to use the "wrong version" kick reason, I thought that got the message across

Fran6nd:
> When I was mulling over this idea I decided to use the "wrong version" kick reason, I thought that got the message across

Ahah thank you. I do not know how I didn't think about it 😅 btw extension should have 3 states in config: enabled, disabled,required.

I am fixing those

piqueserver/config/config.toml
```diff
@@ -85,6 +85,20 @@ scripts = [
...
+#   "enabled"  - server advertises it; clients may use it.
+#   "disabled" - server doesn't advertise it.
+#   "required" - server advertises it and kicks clients that don't
+#                support it back. clients with the "kick reason"
+#                extension see a descriptive message; others get the
+#                generic wrong-version dialog.
+proto_extensions = [
+  ["chat type", "enabled"],
+  ["kick reason", "enabled"],
+]
```
NotAFile: So my personal critique of this approach is that I think this gets confusing very quickly when you have this done via the config. I don't think it's wise to make users configure such a low level feature of a protocol in the config.

I think there's a few categories extensions fall into:
- Non-essential quality of life stuff, e.g  Kick Reason: I don't think it makes sense to have these be configurable at all. They should just always be enabled opportunistically, if the client supports it. There's no reason to disable these, so it's just more boilerplate in the config to keep up to date.
- Extensions required to make an optional server feature work. This gets confusing fast as admins would have to keep track of which exact protocol extensions are required for these features to work and keep those up to date. I think these extensions should just be enabled automatically if the corresponding server feature is enabled. The difficulty there is that it may not always be obvious whether your server configuration is compatible with certain clients.
- Extensions required as part of a specific server configuration. The easy case here is game modes that rely on specific protocol extensions. In that case the responsibility of enabling or requiring those extensions should rest with the gamemode script. The more complicated case is servers like aloha that ban old client versions to ensure a shared baseline across players. I think this is the one case where configuration via the config file is worthwhile, but I'm not sure how common this actually is.

So here's what my personal approach would look like:
- no configuration in the config file, for now
- a script api that allows you to set an extension as enabled/disabled/required, with a text field for the reason it is enabled/disabled/required. This would also be used internally for the default-enabled extensions to keep things consistent.
- print the required feature and reason it is enabled on kick for easy debugging

Fran6nd: Hi @NotAFile ,
Thank you for the review. I did share it on aloha and everyone shares your view.

I Will reimplement it this way.

piqueserver/config/config.toml
```diff
@@ -85,6 +85,8 @@ scripts = [
 # number of /login attempts allowed before players are auto-kicked (default: 1)
 login_retries = 3
 
+# per-extension state for the protocol extensions piqueserver knows about.
```
NotAFile: missed a comment :)

pyspades/player.py
```diff
@@ -144,6 +145,82 @@ def on_ext_info_received(self, contained: loaders.ProtocolExtensionInfo) -> None
...
+        for ext_id, min_ver, reason, name in missing:
+            client_ver = self.proto_extensions.get(ext_id)
+            version_note = ""
+            if client_ver is not None:
+                version_note = " (your client: v{}, needed: v{})".format(
+                    client_ver, min_ver)
+            log.info("{player}: missing enabled extension {name!r}{ver} ({reason})",
+                     player=self, name=name, ver=version_note, reason=reason)
+            self.send_chat_warning(
+                'Heads up: this server uses the "{name}" protocol '
```
NotAFile: I like this feature, I wonder though, how many people still use/prefer voxlap? Will they be annoyed by messages for a client that they can't upgade? I think it's probably worth doing regardless but I just don't know what the situation there is.

Also, iirc these warning type messages are displayed as a toast in openspades, are they readable in that context?

pyspades/player.py
```diff
@@ -144,6 +145,82 @@ def on_ext_info_received(self, contained: loaders.ProtocolExtensionInfo) -> None
...
+                'extension ({reason}). Your client lacks it{ver}, so '
+                "related features won't work for you. Please update "
+                "your client, or open a ticket with its authors to add "
```
NotAFile: I would make this message shorter, but i'll send a proper suggestion

pyspades/proto_extensions.py
```diff
@@ -0,0 +1,136 @@
+# Copyright (c) Mathias Kaerlev 2011-2012.
```
NotAFile: Author and date is wrong, you can put yourself here along with "Piqueserver Authors" if you want

NotAFile (commented):
This is lovely, thanks so much! 💚

pyspades/player.py
Fran6nd: waiting for your input :)

pyspades/proto_extensions.py
Fran6nd: Nice I did not know what to do thank you.

pyspades/player.py
Fran6nd: - About the toast part, I it is visible but in case of multiple messages at the same time only the last one is visible so I am moving toward chat messages ;)
- About Voxlap, as it is unmaintainable, I do not see any smoother way to tell users to migrate to a maintained client.

Fran6nd:
I squashed my commits by mistake while tryin to import stuff from my dev branch... I will try to fix it 🙈 BTW kicking clients with a reason is a nightmare as their client must have started to listen from the chat to receive the reason. Tryin to fix this too.

NotAFile:
> BTW kicking clients with a reason is a nightmare as their client must have started to listen from the chat to receive the reason. Tryin to fix this too.

ohh, not sure if anyone else suggested this but that wasn't what I'd personally meant. It seems very useful to do if we can pull it off, but I mostly wanted it in the server logs so admins would know. Telling clients is gonna be very difficult yeah.

NotAFile:
I think on the kick reason thing, we should just implement it "correctly", since compatability with old clients isn't as much of a concern anyway? Then client developers will hopefully fix their clients to display the kick reason to players.

a7u8:
> I squashed my commits by mistake while tryin to import stuff from my dev branch... I will try to fix it

If you want to checkout the commit you had 2 force-pushes ago, `git checkout 740bc109716d2b58a9f3e23a147e130f2123b381`
If you want to undo all changes to history made after 2 force-pushes ago, but keep them in your working tree, `git reset --soft 740bc109716d2b58a9f3e23a147e130f2123b381` (or remove `--soft` if you don't want to stage them)
If you want to operate on a different commit, see `git reflog`
If someone cloned your repo before you destroyed history, pull from that repo
If your computer melts down and your reflog is lost, all commits pushed to a github fork network are eternal (740bc109716d2b58a9f3e23a147e130f2123b381) and accessable from any fork in the network via the web interface (see also the parent field tucked away below the commit message -- 55cca578071340a461ab7be2c9b7cd713d0ba939); this is a "feature", not a "major security issue"

Fran6nd:
Thanks @DavidCo113 I am on it. @NotAFile for current implementation it will kick with the "Wrong protocol version".
I agree that a proper implementation would be better. Current one is more a hack than anything else. How would you see it? a version 2 of kick reason protocol extension with a dedicated packet this time? Followed by a simple disconnection?

Fran6nd:
Or Clients have the responsibility to tell the user the extension handshake details report when the "Wrong protocol version" kick occurs? Then nothing server side?

a7u8:
> Or Clients have the responsibility to tell the user the extension handshake details report when the "Wrong protocol version" kick occurs? Then nothing server side?

If clients could report that much then they could probably just as easily support the custom kicks instead.

> hi guys :slightly_smiling_face: especially totally_not_a_burner for your knowledge of the protocol... On piqueserver we have an issue with kickreason beeing "not enough". The reason cannot be sent if the client did not spawn first. Maybe time for a version 2? i.e. with a dedicated packet?

Why's that? The betterspades-1-second-thing? Why can't you just extend support of the existing solution?

a7u8:
Digging through your commits

> OpenSpades (and probably other clients) discard ChatMessage packets
received during the initial handshake / map-loading phase, so the kick
reason sent right before disconnect at on_ext_info_received never
surfaces in the disconnect dialog -- the user only sees the generic
"You were kicked from this server" text.

Here's a relevant hack I made a while ago, in its raw and untouched state. It waits for 1 second to handle betterspades and sends the smallest possible """valid""" map down the wire (or would if that `#if 0` was changed to `#if 1`, anyway):
```c
#include <enet/enet.h>
#include "libspades/common/clock.h"
#include <stdio.h>

/* Connect on aos://16777343:32876 */

int main(void) {
	ENetAddress addr;
	ENetHost *host;

	addr.host = ENET_HOST_ANY;
	addr.port = 32876;
	
	host = enet_host_create(&addr, 32, 1, 0, 0);
	if (host == NULL) {fputs("ENet host creation is not working. You're not using port 32876, right?\n", stderr);}

	enet_host_compress_with_range_coder(host);

	while (1) {
		ENetEvent event;

		enet_host_service(host, &event, -1);
		if (event.type == ENET_EVENT_TYPE_CONNECT) {
			uint64_t clock = libspades_get_clock()+LIBSPADES_SECONDS_TO_CLOCK_UNIT(1);
			uint8_t state[84] = {15};
			ENetPacket *packet;

			puts("connect");

			while (1) {
				uint64_t clock2 = libspades_get_clock();
				if (clock2 >= clock)
					break;

				enet_host_service(host, NULL, LIBSPADES_CLOCK_UNIT_TO_MILLISECONDS(clock-clock2));
			}

			puts("goodbye");

#if 1
#if 0
			/* Not actually necessary for BetterSpades, or even OpenSpades, but left disabled here anyway. */
			/* As for Voxlap, disabling this gives a map uncompression error. Enabling it gives "Block" as an error. */
			packet = enet_packet_create("\x12\x00\x00\x00\x00", 5, ENET_PACKET_FLAG_RELIABLE);
			enet_peer_send(event.peer, 0, packet);

			packet = enet_packet_create("\x13\x78\xda\x63\xb0\xb3\x01\x00\x00\xbb\x00\x7b", 12, ENET_PACKET_FLAG_RELIABLE);
			enet_peer_send(event.peer, 0, packet);
#endif

			/* Technically optional but prevents BetterSpades from showing a completely uninitialized map. */
			state[0] = 15;
			packet = enet_packet_create(state, 84, ENET_PACKET_FLAG_RELIABLE);
			enet_peer_send(event.peer, 0, packet);
#endif

			packet = enet_packet_create("\x11\xff\x02reason", sizeof("\x11\xff\x02reason"), ENET_PACKET_FLAG_RELIABLE);
			enet_peer_send(event.peer, 0, packet);

			/* Banned. */
			enet_peer_disconnect_later(event.peer, 1);
		}
	}
}
```

NotAFile:
What I meant is that currently maintained clients should be updated to not ignore the kick reason chat message. The rest will just see the "wrong version" message which is fine.

pyspades/player.py
```diff
@@ -1324,6 +1327,20 @@ def send_map(self, data: Optional[ProgressiveMapGenerator] = None) -> None:
...
+            # state machine doesn't honor them) and the kick-reason text
+            # never surfaces in the disconnect dialog. Worse, queueing
+            # the kick chat + disconnect_later immediately tends to make
+            # the peer time out before its queue can drain, so the client
+            # ends up seeing a generic data=0 disconnect.
+            #
+            # A short delay lets the client receive StateData, run
+            # MapLoaded(), transition to NetClientStatusConnected, and
+            # start handling chat. The kick reason then renders properly.
+            reactor.callLater(1.0, self._enforce_mandatory_extensions)
```
NotAFile: This delay amount is just a best guess, right? Is there any easy way we can send it exactly when it's needed?

Fran6nd: It is now outdated I will publish a comment explaining how it is now ;)

Fran6nd:
The 1.0s callLater is now replaced by `_wait_for_map_ack_then_kick` in `pyspades/player.py`, which waits on a real signal instead of a fixed guess.

How it works now

After send_map queues StateData (the packet that flips the client into NetClientStatusConnected) and we've called on_join, we poll peer.reliableDataInTransit every 50ms. That's ENET's count of un-ACK'd reliable bytes; when it hits 0, the client's networking layer has received everything we sent. So the wait scales with RTT instead of being a flat second.

A 5s hard ceiling guards against misbehaving or unreachable peers — on timeout we fall back to a plain disconnect(ERROR_KICKED) with no kick-reason chat.

The remaining 0.3s

Once reliableDataInTransit hits 0 we still wait an extra 0.3s before sending the kick chat. ENET emits ACKs from inside enet_host_service before the receiving app dispatches the packet, so the ACK only proves the client's networking layer has the bytes — not that MapLoaded() has finished and the client is in NetClientStatusConnected. Without that cushion the kick chat sometimes arrives mid-load and the dialog reads "Unknown reason" again. 0.3s covers one or two client frames at any plausible framerate.

Total wait: (RTT-bounded ACK time) + roughly 300ms instead of a flat 1s. The only way to drop the 0.3s server-side would be for the client to send an explicit "ready" packet from inside MapLoaded() — noted as the long-term fix in the docstring, but it needs a client-side change.

This is the cleanest hack/solution I got which does not need any changes client side. What are your thoughts?

Tested on ZeroSpades: The client hear the song but never sees the menu asking to select a team 😉

Fran6nd:
The pros of this is backward compatibility. However if a v2 of kick reason extension is successfully defined, as a maintainer of zerospades I am willing to implement it. I believe that it will be needed sooner or later.

Fran6nd:
Capture d’écran 2026-05-19 à 17 22 19

This is the message received by the client from ZeroSpades perspective. You can see extension names.


If you wanna test, here is my dummy protocol extension script I use for testing:

```
from pyspades.constants import EXTENSION_NAMES

# ids no real client implements -- guarantees the "missing" branch fires.

# ENABLED dummies: client connects, sees one chat line per entry plus the
# single call-to-action line.
_FAKE_ENABLED = [
    (248, "test feature A",
     "test toast: short reason",
     1),
    (249, "test feature B",
     "test toast: pretend this is a competitive-mode HUD addition",
     1),
    (250, "test feature C",
     "test toast: pretend this is a new piqueserver HUD extension that "
     "benefits players who run a modern client",
     1),
]

# MANDATORY dummies: client connects, gets kicked. With both listed the
# kick reason should mention both. (Note: kick reason is truncated to
# MAX_CHAT_SIZE = 90 chars on the wire, so very long reasons may clip.)
_FAKE_MANDATORY = [
    (246, "mandatory A", "required HUD", 1),
    (247, "mandatory B", "required scoring",  1),
]

# make _name() in the registry print readable strings instead of "id NNN"
for _ext_id, _name, _reason, _version in _FAKE_ENABLED + _FAKE_MANDATORY:
    EXTENSION_NAMES[_ext_id] = _name


    class WarnToastTestProtocol(protocol):
        def __init__(self, *args, **kwargs):
            super().__init__(*args, **kwargs)
            for ext_id, name, reason, version in _FAKE_ENABLED:
                self.extensions.enable(ext_id, reason=reason,
                                       version=version)
            for ext_id, name, reason, version in _FAKE_MANDATORY:
                self.extensions.mandate(ext_id, reason=reason,
                                        version=version)
            print(
                "[_warn_toast_test] enabled: {}".format(
                    ", ".join("id={} ({!r}) v{}".format(eid, n, v)
                              for eid, n, _, v in _FAKE_ENABLED)))
            print(
                "[_warn_toast_test] mandatory: {}".format(
                    ", ".join("id={} ({!r}) v{}".format(eid, n, v)
                              for eid, n, _, v in _FAKE_MANDATORY)))

    return WarnToastTestProtocol, connection

```

pyspades/player.py
a7u8: <a href="https://raw.githubusercontent.com/xtreme8000/BetterSpades/refs/heads/standalone/src/network.c#:~:text=listen%20connection%20for%201s">Nope!</a>

Fran6nd:
It seems that my latest idea is breaking BetterSpades (makes it crash).
So here is my take:
- We send wrong version disconnection for all clients supporting kick reason v1 or not supporting it.
- We prepare a kick reason v2 that support kicking from the the beginning of the enet connection. All client advertising it will be aware of that. V2 could be only pure semantic: Same as v1, no new packet, but each client advertising v2 MUST be able to receive kick reason AT ANY TIME.

So we will explicitly know if client supports it, and avoid client specific hacks. What do you think?

EDIT: if we really want it I should be able to provide client dependent hack but I cannot recommend that.

Fran6nd:
https://github.com/Fran6nd/aosprotocol/blob/kickreasonV2/extensions.md

I did document here what a v2 would be. I can add it to the [ongoing PR](https://github.com/piqueserver/aosprotocol/pull/54) I have that documents other extensions?

And here I can provide commits implementing it?

EDIT: I have a zerospades branch with v2 implemented [here](https://github.com/Fran6nd/zerospades/pull/new/kickreasonV2) and in this PR I implemented it too for discussion. If needed I can of course strip it. The cost of V2 is really minimalistic and need almost no work at all server side while remaining fairly easy client side. And harmonizing client behaviors cannot be bad.

EDIT 2: We could, for client not supporting V2 BUT supporting V1 wait for them like 1 second before kicking them with a reason (terrible ux but at least they will know and it would work for better spades and openspades)

a7u8:
```
ENABLED = "enabled"
MANDATORY = "mandatory"
DISABLED = "disabled"
```

There's no indication that these are specific to proto_extensions; why not just pass in the actual underlying string values directly, anyway? Or use an enum, or possibly an opaque `object()` for each of them.

Fran6nd:
Finally I did not bump kick reason to v2 and fixed V1 behavior.
https://github.com/zerospades/zerospades/pull/49
To my knowledge kyrospades fixed it the same way too.
I will update this PR accordingly.

PR #850, opened 2026-05-25
Add support for the Player Properties extension.

Fran6nd:
This PR is needed to have pique comply with existing protocol extension (There is still utf-8 extension missing to my knowledge).

The spec of this extension is not from me but in [my ongoing PR ](https://github.com/piqueserver/aosprotocol/pull/54) I am documenting it. Be aware that zerospades already supported it (with [an issue ](https://github.com/zerospades/zerospades/pull/35)about score I discovered while testing this PR )for a long time, as well as better spades and libspades.

The server does not need it nor use it to run. The extension is available for game mode to use if needed.

Worth noting that my PR does not allow sending the data of a player to another player. As to my knowledge, clients are not aware of other player's hp/ammo... Somehow score would make sense but that should be for an hypothetic future sub packet id ?

### DIFF
--- a/piqueserver/server.py
+++ b/piqueserver/server.py
@@ -57,7 +57,8 @@ from piqueserver.utils import as_deferred, EndCall
 from piqueserver.bansubscribe import bans_config_urls
 from pyspades.bytes import NoDataLeft
 from pyspades.constants import (CTF_MODE, ERROR_SHUTDOWN, TC_MODE,
-                                EXTENSION_CHATTYPE, EXTENSION_KICKREASON)
+                                EXTENSION_CHATTYPE, EXTENSION_KICKREASON,
+                                EXTENSION_PLAYERPROPERTIES)
 from pyspades.master import MAX_SERVER_NAME_SIZE
 from pyspades.server import ServerProtocol, Team
 from pyspades.tools import make_server_identifier
@@ -277,6 +278,7 @@ class FeatureProtocol(ServerProtocol):
         self.bans = NetworkDict()
 
         self.available_proto_extensions = [
+            (EXTENSION_PLAYERPROPERTIES, 1),
             (EXTENSION_CHATTYPE, 1),
             (EXTENSION_KICKREASON, 1),
         ]
--- a/pyspades/constants.py
+++ b/pyspades/constants.py
@@ -74,15 +74,14 @@ ERROR_SHUTDOWN = 5
 ERROR_KICKED = 10
 ERROR_INVALID_NAME = 20
 
+# Base packet id for protocol-extension packets. Each extension's
+# packet id is PACKET_EXT_BASE + ext_id.
+PACKET_EXT_BASE = 0x40
+
 EXTENSION_PLAYERPROPERTIES = 0
 EXTENSION_PLAYERLIMIT = 192
 EXTENSION_CHATTYPE = 193
 EXTENSION_KICKREASON = 194
-
-# Sub-ids for the Player Properties extension (packet id 64). Sub id 0 is
-# the only one currently defined; preserved on the wire so future sub-ids
-# can be dispatched by callers.
-PLAYERPROPERTIES_SUB_QUERY = 0
 
 CTF_MODE = 0
 TC_MODE = 1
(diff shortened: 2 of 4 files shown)


NotAFile:
One might consider occasionally sending this to the client to reduce desyncs but those types of desyncs haven't really ever been an issue to my knowledge?

Fran6nd:
> One might consider occasionally sending this to the client to reduce desyncs but those types of desyncs haven't really ever been an issue to my knowledge?

Not to mine too :) i never faced the issue.

a7u8:
If you can you should probably make it compatible with 0752b6d1abfa1962973df936f33db966f7496c4a

> (There is still utf-8 extension missing to my knowledge).

UTF-8 extension has no ID, it predates the whole betterspades--piqueserver extension packetry by a long time -- piqueserver "supports" it already if you ignore the completely broken CP-437 codec silently converting codepoints <0x20 from their normal meaning into CP-437 pictograms.

> Worth noting that my PR does not allow sending the data of a player to another player. As to my knowledge, clients are not aware of other player's hp/ammo...

BetterSpades tries to estimate the amount of ammunition another player has left and will forcibly stop that player from visibly shooting when it reaches 0. This has no tangible purpose other than to cause desync. Possibly it's a workaround for BetterSpades' bug with always sending WeaponInput as if it's mouse input and not actual trigger input to the gun.

Sending PlayerProperties is the only effective way I know of to work around that bug.

It could also be useful for spectating too, if a client knew how to handle the packet (I *think* betterspades can display HP and ammo).

> Somehow score would make sense but that should be for an hypothetic future sub packet id?

Why make a whole new packet for it without supporting the old one?

Fran6nd:
Thank you for all these informations. I am learning a lot. I did not know about Betterspades's issue. So being able to broadcast player properties makes sense. @NotAFile do you want this PR to feature an integration into the pique server codebase?

I think it could be a separate PR from the base implementation?

I will also update my PR documenting the extension with the new info learned here.

> It could also be useful for spectating too, if a client knew how to handle the packet (I _think_ betterspades can display HP and ammo).

Hell yeah. and for demo replaying! Being able to display the HUD would be sooo cool.

> Why make a whole new packet for it without supporting the old one?
Ahah as I did not understand why client would know the precise status of others, in openspades variants, the only visible variable was score. Idea was To set the score of a particular client to all without having set all the fields. With the new understanding I have it is useless though.


I will update my PR thank you !

NotAFile:
> do you want this PR to feature an integration into the pique server codebase? I think it could be a separate PR from the base implementation?

Whatever you feel like! I don't mind.

One other thing I would like to avoid if possible in the integration is sending more precise info on things like health/ammo to other players since that could be surfaced by a cheat. Information should be limited to the minimum necessary for clients to display the game correctly.

Fran6nd:
> > do you want this PR to feature an integration into the pique server codebase? I think it could be a separate PR from the base implementation?
> 
> Whatever you feel like! I don't mind.
> 
> One other thing I would like to avoid if possible in the integration is sending more precise info on things like health/ammo to other players since that could be surfaced by a cheat. Information should be limited to the minimum necessary for clients to display the game correctly.

So a method is now provided to broadcast a player properties but unused in pique server for now.

I agree however if I understood correctly @DavidCo113  to avoid desync in the future we will have no choice other than broadcasting player properties? for better spades.

Anyway, this PR will feature only implementation but no usages yet of it. PlayerProperties will be available for scripts.

To me it is ready for review. I can squash commits if you want.

a7u8:
> I agree however if I understood correctly @DavidCo113 to avoid desync in the future we will have no choice other than broadcasting player properties? for better spades.

Yep, but not too often; just set the ammo of other players to 255 every once in a while.
SMG fires at 10 Hz, so 25.5 seconds is the most you could go without sending anything on a perfect connection (probably actually send it more often than that). In exchange for a small amount of complexity you could also only bother sending if the player's fired recently.

NotAFile:
It feels like fixing this on the betterspades side (or working with them) would be much easier, but yeah, in that case I would recommend sending dummy values. But that can wait, this is already useful as is.

